PR #7, opened 2020-11-23
Preliminary CI

philass:
I had the most barebones CI. There are not a lot of tests right now so its not that meaningful. But as more tests are written. It can be pretty convenient to see if PRs and commits break.

Let me know if there are any changes you want me to make, and I more than happy to :). I'll also work to build on this in the future to make it better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: rust
+sudo: false # <-- always goodness
+cache: cargo
+rust:
+  - nightly
+matrix:
+  fast_finish: true
+script:
+  - cargo build --verbose --all
+  - cargo test --verbose --all


bkolobara:
Thanks for taking the time to contribute to Lunatic.

This uses TravisCI. How does this compare to other CIs, like Github Actions?
CI will be an important part of the development process and I would like to weight all pros and cons of different solutions before settling for one. Getting it right at the beginning will save us some time later.

It's also unclear to me how this would work. I assume I also need to sign up for an account at https://travis-ci.org/ and configure something there? Could you write some instructions here how to use it?

Also, I can't figure out their pricing structure, how much would this approximately cost per month to test/build on Linux, MacOs (x64 and ARM) and Windows?

PR #8, opened 2020-11-25
Preliminary CI

philass:
Making the PR again

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jobs:
       script: echo "Deploying to GitHub releases ..."
       deploy:
         provider: releases
-        api_key: $GITHUB_OAUTH_TOKEN
+        # api_key: $GITHUB_OAUTH_TOKEN
         skip_cleanup: true
         on:
           tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,11 @@ matrix:
 script:
   - cargo build --verbose --all
   - cargo test --verbose --all
+- stage: GitHub Release
+  script: echo "Deploying to GitHub releases ..."
+  deploy:
+    provider: releases
+    api_key: $GITHUB_OAUTH_TOKEN
+    skip_cleanup: true
+    on:
+      tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: rust
+sudo: false # <-- always goodness
+cache: cargo
+rust:
+  - nightly
+matrix:
+  fast_finish: true
+script:
+  - cargo build --verbose --all
+  - cargo test --verbose --all
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,18 @@ sudo: false # <-- always goodness
 cache: cargo
 rust:
   - nightly
-matrix:
-  fast_finish: true
-script:
-  - cargo build --verbose --all
-  - cargo test --verbose --all
-- stage: GitHub Release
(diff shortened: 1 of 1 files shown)


philass:
Answering the questions from the previous PR


"This uses TravisCI. How does this compare to other CIs, like Github Actions?
CI will be an important part of the development process and I would like to weight all pros and cons of different solutions before settling for one. Getting it right at the beginning will save us some time later."

Github Actions seems to be the most feature rich. Travis does seem to be the simplest though, and the easiest to get going with in my experience. I also see it used in a lot of Rust repositories. However I don't know if they use the free tier or not. But I'll look into github actions some more.

It's also unclear to me how this would work. I assume I also need to sign up for an account at https://travis-ci.org/ and configure something there? Could you write some instructions here how to use it?

So all you have to do is go to travis-ci.com (not .org... they are phasing that out). Once you are there you simply have to give grant permission for Lunatic. And it will automatically run the build if there is a .travis.yml present. 

Also, I can't figure out their pricing structure, how much would this approximately cost per month to test/build on Linux, MacOs (x64 and ARM) and Windows?

I think with Github Actions this would actually be free for all 3 options. With Travis linux and Windows are free (up to 1000 mins of build time a month), where as MacOs is not. This probably means that github actions is a better choice for Lunatic.

bkolobara:
Thanks for taking the time and researching this! You are right, when I go to the .com site (instead of .org) they have more information.

According to their [pricing page](https://travis-ci.com/plans) and [this blog post](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) they don't have a free plan. Only a free trial period that include 10k credits, but once you spend it you need to upgrade to a paid one. You also need to pay something extra for MacOs, even the pricing page "contradicts" the blog post in this regard.

I don't want to focus too much on pricing, as both options seem reasonable. Looking at the documentation TravisCI seems indeed a bit simpler to work with and more straight forward. I just feel that Github Actions would be tightly integrated with the rest of Github and maybe more future proof.

I would stick for now with TravisCI. The cost of switching to something else later if we run into problems doesn't seem too much.

As a first step, could you add to this PR the functionality of creating releases just by pushing new tags as described here: https://docs.travis-ci.com/user/deployment/releases/ ?

I guess, I would then merge this and try it out? Is there maybe another way of trying out the `.travis.yml` file, without committing it to repository? Committing every iteration until I get the configuration right seems a bit unnecessary.

grippy:
One thing I would consider adding here is having the CI use Clippy (or something equivalent) for a few things:
- checking if `cargo fmt` needs to applied
- adding lint checks for dealing rejecting default compiler warnings (code, tests, and documentation).

You can probably start with this?
https://github.com/rust-lang/rust-clippy#travis-ci

Ideally, there would be a reference in Travis config to either a script which wraps all these clippy checks (or a Makefile target). Then you would run locally prior to submitting your PR.

Just food for thought.

philass:
I think that makes a lot of sense. It seems that people's trust in Travis CI is going quickly, with their constantly changing policies on free/paid tiers. I think we should probably target these things with github actions. They seem to have a more comprehensive free tier. But I'll look into seeing how to implement the clippy equivalent in github actions.

bkolobara:
I have decided to go ahead with Github Actions, it seems like a safe bet right now. I already added a workflow to automatically generate release builds when a new tag is added. I have played a bit around with it and it looks pretty nice to me.

I will be progressively adding other checks (formatting & clippy).

Thanks a lot @philass for starting the discussion and giving some feedback on this.

Issue #11, opened 2020-12-03
unresolved import `lunatic::net`


ArneVogel:
Hi,

I was trying out the networking example from [this](https://dev.to/bkolobara/writing-rust-the-elixir-way-2lm8) blog post and encountered this error:
```
error[E0432]: unresolved import `lunatic::net`
 --> src/main.rs:1:24
  |
1 | use lunatic::{Process, net}; // Once WASI gets networking support you will be able to use Rust's `std::net::TcpStream` instead.
  |                        ^^^ no `net` in the root
```

Cargo.toml:
```
[dependencies]
lunatic = "0.2.0"
```
Building with `cargo build --release --target=wasm32-wasi`

Thanks,
Arne

bkolobara:
Hi @ArneVogel! Thank you for reporting this.

The issue is that I used the source code directly as a dependency for the blog post instead of `lunatic = "0.2.0"`.
I have a published the `0.2.1` release containing the latest source, it should work now.

PR #12, opened 2020-12-03
re-export net

ArneVogel:
Fixes issue #11.

Compiling the code from the networking example from this [blog post](https://dev.to/bkolobara/writing-rust-the-elixir-way-2lm8) with `lunatic = { git = "https://github.com/ArneVogel/lunatic", branch = "net" }` instead of `lunatic = "0.2.0"` results in the expected behavior.

### DIFF
--- a/stdlib/rust/src/lib.rs
+++ b/stdlib/rust/src/lib.rs
@@ -49,6 +49,7 @@ pub mod net;
 pub mod process;
 
 pub use channel::Channel;
+pub use net::{TcpListener, TcpStream};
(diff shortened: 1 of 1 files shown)


bkolobara:
Hi @ArneVogel. Thanks for taking the time investigating this.

The `lunatic::net` namespace is already public, it was just not part of the `0.2.0` release on crates. I have published now `0.2.1` containing the latest code.

Your example works because it uses the source directly and not because the new top level exports.

PR #16, opened 2020-12-28
Fixed failing patching tests

imor:
The yield function in the lunatic module was recently renamed to yield_,
but the patching tests still expected yield in the patched modules.


bkolobara:
The last commits touched many parts of the code base and I must have missed this. Thanks @imor for the fix!

PR #19, opened 2021-02-03
Unify state wrapper

jocaml:


src/process/mod.rs
```diff
@@ -137,10 +137,10 @@ impl Process {
 
         let stack = OneMbStack::new()?;
         let mut process = AsyncWormhole::new(stack, move |yielder| {
-            let yielder_ptr = &yielder as *const AsyncYielderCast as usize;
+            let yielder_ptr = &yielder as *const AsyncYielder<Result<A::Return>> as usize;
```
bkolobara, line 140: We are casting the yielder back to `AsyncYielder<Result<()>>` here: https://github.com/lunatic-solutions/lunatic/blob/unify_state_wrapper/src/process/mod.rs#L67

This should also be updated to use the new type. I forgot why I couldn't just pass around the pointer as `*const AsyncYielder<Result<A::Return>>` and needed to cast it to `usize`. Maybe this could be refactored a bit so that we don't destroy the type information.

uptown_funk/src/lib.rs
```diff
@@ -121,13 +127,46 @@ impl<S, E: Executor> StateWrapper<S, E> {
...
+    pub fn recover_state(self) -> Result<S, ()> {
+        match Rc::try_unwrap(self.state) {
+            Ok(s) => Ok(s.into_inner()),
+            Err(_) => Err(()),
+        }
+    }
+}
+
+unsafe impl<S, E: Executor> Send for StateWrapper<S, E> {}
+unsafe impl<S, E: Executor> Sync for StateWrapper<S, E> {}
```
bkolobara, line 151: This is only required by Wasmer and can be marked with: `#[cfg(feature = "vm-wasmer")]`.
`StateWrapper` is obviously `!Send` and `!Sync` because of `Rc<RefCell>`;

For now this is ok, but could be an issue later. Not sure what the best approach here is.

PR #26, opened 2021-03-03
Wasi clock_res_get and clock_time_get

akegalj:
This implements `clock_res_get` and `clock_time_get` apis for wasi.

Implementation is based on wasmer.

There are two new dependencies:
 - libc => adds functionality for unix
 - winapi => adds functionality for windows

There are some unimplemented methods for windows platform (same is with wasmer) which I didn't go deep into atm but basic clock should work. Error handling can be a bit better for unix implementation and there are some TODOs for that.

Build passes for both `x86_64-pc-windows-gnu` and default linux target.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -360,16 +360,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cpu-time"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e9e393a7668fe1fad3075085b86c781883000b4ede868f43627b34a87c8b7ded"
-dependencies = [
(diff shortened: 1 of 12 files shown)


jtenner:
Given how sensitive clock time res and get methods are, and how important [spectre](https://en.wikipedia.org/wiki/Spectre_(security_vulnerability)) is... my question is this:

Are we introducing security vulnerabilities with this implementation? I'm definitely not the expert 😅 .

jocaml:
@jtenner also not an expert, but we can add a parameter to specify max resolution. I will merge this and maybe we can do a bit of research on timing attacks and clock resolution and create a "begginer-friendly" issue.

PR #32, opened 2021-03-09
Fix file time metadata

akegalj:


src/api/wasi/state.rs
```diff
@@ -186,7 +205,7 @@ impl FileDesc {
     }
 
     fn open_with_flags<P: AsRef<Path>>(path: P, flags: OpenFlags) -> Result<Self, Status> {
-        if fs::metadata(&path)?.is_dir() {
```
withtypes, line 189: @akegalj how come you changed it? If metadata returns `Err`, shouldn't we propagate it here?

src/api/wasi/state.rs
```diff
@@ -133,24 +143,33 @@ impl WasiState {
 
     pub fn filestat_path<P: AsRef<Path>>(&self, abs_path: P) -> Result<Filestat, Status> {
         let metadata = fs::metadata(&abs_path)?;
+        // NOTE try_into cast should work correctly on glibc systems as tv_nsec will be less
```
withtypes, line 146: @akegalj so `tv_nsec` is `u32` less than 1e9, but even if it was 2^32 - 1, this would work hm They simply note that it's less than 1e9 because that's a whole second. Anyway, I think this is fine.

src/api/wasi/state.rs
withtypes: @akegalj please check this so I can merge

akegalj: > shouldn't we propagate it here?

no, this function is used for opening a file. Lets asume user/guest wants to create a file in a home directory and he does `std::fs::File::create("/home/user/new_dir");`. In this case our `api::wasi::api::WasiState::path_open` will be invoked and soon after `api::wasi::state::WasiState::open_with_flags` which will invoke `fs::metadata("/home/user/new_dir")`.

Unfortunatelly this will error out becase `metadata` errors https://doc.rust-lang.org/std/fs/fn.metadata.html#errors if:
> path does not exist.

In this case it should be valid to create a new file for the user/guest.

In the case `"/home/user/new_dir"` is an existing directory, we will just open the directory:
```
         if let Ok(true) = fs::metadata(&path).map(|f| f.is_dir()) {
             return Self::open(&path);
         }
```
 (this is a workaround because we don't implement flags correctly and https://doc.rust-lang.org/std/fs/fn.read_dir.html `read_dir("/home/akegalj/bla5")` would fail wtih:
 ```
 Error: Os { code: 28, kind: InvalidInput, message: "Invalid argument" }
 ```
 even if the directory was existing. The problem seems to be with `.write(true)` from the next line:
 ```
         let file = OpenOptions::new()
            .read(true)
            .write(true) // ==> this line fails when read_dir is invoked
            .create(flags.create())
            .truncate(flags.truncate())
            .open(&path)?;
```

I didn't go deeper why `read_dir` fails when `write(true)` is set (which by the way makes sense; read_dir should fail if `write` flag is true and this is probably the reason) and line:
```
if let Ok(true) = fs::metadata(&path).map(|f| f.is_dir()) {
```
is a workaround for that case.

Conclusion is that:
```
if fs::metadata(&path)?.is_dir() {
```
fails on at least one corner case (`read_dir`) than:
```
if let Ok(true) = fs::metadata(&path).map(|f| f.is_dir()) {
```
.

The most correct solution to this seems to be:
```
    fn open_with_flags<P: AsRef<Path>>(path: P, flags: OpenFlags) -> Result<Sel
f, Status> {
        // if case removed!
        let file = OpenOptions::new()
            .read(true)
            .write(flags.create())    // ==> this changed!
            .create(flags.create())
            .truncate(flags.truncate())
            .open(&path)?;
        let path = PathBuf::from(path.as_ref());
        Ok(Self { file, path })
}
```
which works for `File::create`, `fs::create_dir` and with `read_dir`. There might be cases where it fails though but it is the solution I would go with if my understanding is right.

I will remove the if clause altogether.

PR #42, opened 2021-04-29
Support mutex/none as state wrappers in host functions

withtypes:
Host functions macro now has additional parameter `sync` which can be set to `none` (previous behavior) or `mutex`. When set to `mutex`, the state will be wrapped into `Arc<Mutex<S>>` and it can be observed from outside while the process is running.


bkolobara:
I ran a few tests and benchmarks on my MacBook and everything looks good to me. Could you just resolve the conflict so that I can merge it?

PR #44, opened 2021-05-04
Add simple heap profiler

akegalj:
# Abstract

This PR adds support for simple heap profiler. If flag `-p` is given, lunatic will output `heap.dat` file that contains profile data (time/heap data points). If more subprocesses are run within profiled process then profile will contain merged data points of master process with child processes.

Tested languages: C, Rust

# Details

Every child process will create its own `HeapProfilerState`. This object tracks history of all invocations of `malloc`, `calloc`, `realloc`, `free`, `aligned_alloc`. History for every process is tracked in ringbuffer (max ~3mb). When child process terminates, his history is merged with parent process history. When parent process terminates, history is written to `heap.dat` file. 

# Caveat

Merging of histories is done in `O((N+M)*log(N+M))` where `N` is parent process history and `M` is child process history and its using `N+M` extra temporary memory. This is to properly implement ringbuffer (older elements get dropped). In practice this means that if we are profiling and child process terminates we are going to temporary alocate ~5mb (2* history buffer) and spend some time to rearange ringbuffer. Benefit is that profiling of long running processes (with many spawned child processes) will be realativelly cheap. At any point in time profiling will be expensive as `N`*`history`, where `N` is processes currently running and `history` buffer with maximum size of ~3mb. History buffer size can be configured.

With little effort (20 loc) the above can be optimized to use `O(N+M)` but atm I don't think its worth the effort (as code becomes less maintainable). There is an extra TODO in code when this would be needed

Objective when designing solution for this feature was that profiling will be used in lunatic cli for development/debugging of guest programs; not in production so simple design/implementation was prefered.

# Alternatives

There are few alternative solutions explored:
 1) defer history merging after master process exists
      - instead of merging immediatelly when child process exists, we could carry all childs heap buffers until master process finishes and then merge.
   * Pros:
        - merging is done after master process terminates so doesn't effect process performance
   * Cons:
        - higher memory footprint for multiprocess apps (because all history buffers are held until master process exits and not merged/cleaned immediatelly)
        
 2) all subprocesses share same `HeapProfilerState`.
   * Pros:
        - low profiling memory overhead (only one history ringbuffer); <3mb memory
        - no merging histories in multiprocess application (no time penalty for merging)
        - enables live profiling (we could observe profiler while program is executing)
   * Cons:
        - time penalty for mutex syncing in multiprocess heavy aplications (ie. only one process at a time can call `malloc`)
        - more complicated tracking of proces pid-s
            * this is due to a bug where `free` is not called for all resources when process is finished. To workaround, we have to manually call `free` for resources traced by some process. This is easy if every process has its own `HeapProfilerState` (as implemented in this PR), but gets trickier if all processes share same `HeapProfilerState`. Then every process should actually call `free(pid, ptr)`, where `pid` is extra argument carying process id. Or `HeapProfilerState` should have access to current process id somehow. This involves extra hacking
        
Also note that in current implementation `HeapProfilerState` is passed from child to parrent process by process return value (in host): https://github.com/lunatic-solutions/lunatic/compare/heap_profiler?expand=1#diff-89815f1e2909a7ef216763b9b930a308e2b718bf17ebfc270433d4abde0635f5R46 . Initially I wasn't happy about this, but actually it makes perfect sense and intention is clear in the source code. Alternitively this state passing from child process to parrent process can be hidden (handled by a mutex/ref passed through all processes) but for this use cases I prefer it as it is done in this PR.

# Extra

Additional flag `-n` is added for easier debugging. This flag will output file `normalisation.wasm` that contains patched/normalised input file.

# Todo

 * There are some TODO's left in comments.
 * normalisation tests could be made simpler
     - tests are run after all patches are applied so reduction counter logic/patch makes heap profiler tests unnecessary verbose. We could refactor `tests/normalisation_patching_test.rs` so that it runs tests for specific patch only. For example reduction counter tests would test only reduction counter patch, and heap profiler tests would test only heap profiler patch. Atm this isn't the case and every test is testing all patches. This has its cons and pros

### DIFF
--- a/src/api/heap_profiler.rs
+++ b/src/api/heap_profiler.rs
@@ -1,0 +1,142 @@
+use std::collections::{HashMap, VecDeque};
+use std::fs::File;
+use std::io::Write;
+use std::time::{Duration, SystemTime};
+use uptown_funk::host_functions;
+
+use log::debug;
+use log::error;
+
+type Ptr = u32;
+type Size = u32;
+const HISTORY_CAPACITY: usize = 100_000;
+
+pub struct HeapProfilerState {
+    memory: HashMap<Ptr, Size>,
+    started: SystemTime,
+    heap_history: VecDeque<(i32, Duration)>,
+}
+
+impl HeapProfilerState {
+    pub fn new() -> Self {
+        Self {
+            memory: HashMap::new(),
+            started: SystemTime::now(),
+            heap_history: VecDeque::new(),
+        }
+    }
+
+    // Merge child process profile results into parent process.
+    //
+    // Usually this should be called when child process exits.
+    pub fn merge(&mut self, profiler: HeapProfilerState) {
+        self.started = std::cmp::min(self.started, profiler.started);
+        let started_delta = profiler.started.duration_since(self.started).unwrap();
+        // TODO sorted state merging of two sorted lists can be done in O(N+M)
+        // currently this runs in O((N+M)*log(N+M)*C)
+        profiler.heap_history.iter().for_each(|(h, d)| {
+            self.heap_history.push_back((*h, *d + started_delta));
+        });
+        self.heap_history
+            .make_contiguous()
+            .sort_unstable_by_key(|&(_, d)| d);
+        // remove oldest extra elements (ringbuffer)
+        if self.heap_history.len() > HISTORY_CAPACITY {
+            self.heap_history
+                .drain(0..(self.heap_history.len() - HISTORY_CAPACITY));
+        }
+    }
+
+    // Deallocate remaining memory. This is invoked when process exits
+    pub fn free_all(&mut self) {
+        // implicitly deallocate all left over child memory
+        self.memory.clone().iter().for_each(|(k, _)| {
+            self.free_profiler(*k);
+        });
+    }
+
+    // Write heap profile history to a file. Format:
+    //
+    // #time/sec heap/byte
+    // 0.1       10
+    // 0.2       15
+    // ...
+    pub fn write_dat(&self, fd: &mut File) -> std::io::Result<()> {
+        let mut graph = Vec::new();
+        let mut heap_size: u64 = 0;
+        writeln!(&mut graph, "#time/sec heap/byte")?;
+        // write initial entry for plotting
+        writeln!(&mut graph, "0 0").unwrap();
+        self.heap_history.iter().for_each(|(heap, duration)| {
+            if *heap < 0 {
+                // TODO check for overflow
+                heap_size -= (-*heap) as u64;
+            } else {
+                heap_size += *heap as u64;
+            }
+            writeln!(&mut graph, "{} {}", duration.as_secs_f64(), heap_size).unwrap();
+        });
+        fd.write_all(&graph)
+    }
+
+    fn history_push(&mut self, change: i32) {
+        if self.heap_history.len() == HISTORY_CAPACITY {
+            // if HISTRY_CAPACITY > 0 this should be safe
+            self.heap_history.pop_front().unwrap();
+        }
+        // TODO: trap/log error if elapsed failed
+        // very unlikely to happen!
+        self.heap_history
+            .push_back((change, self.started.elapsed().unwrap()));
+    }
+}
+
+#[host_functions(namespace = "heap_profiler")]
+impl HeapProfilerState {
+    fn aligned_alloc_profiler(&mut self, _self: u32, size: Size, ptr: Ptr) {
+        self.malloc_profiler(size, ptr);
+    }
+
+    fn malloc_profiler(&mut self, size: Size, ptr: Ptr) {
+        debug!("heap_profiler: malloc({}) -> {}", size, ptr);
+        self.memory.insert(ptr, size);
+        self.history_push(size as i32);
+    }
+
+    fn calloc_profiler(&mut self, len: Size, elem_size: Size, ptr: Ptr) {
+        debug!("heap_profiler: calloc({},{}) -> {}", len, elem_size, ptr);
+        let size = len * elem_size;
+        self.memory.insert(ptr, size);
+        self.history_push(size as i32);
+    }
+
+    fn realloc_profiler(&mut self, old_ptr: Ptr, size: Size, new_ptr: Ptr) {
+        debug!(
+            "heap_profiler: realloc({},{}) -> {}",
+            old_ptr, size, new_ptr
+        );
+        match self.memory.remove(&old_ptr) {
+            Some(removed_size) => {
+                self.memory.insert(new_ptr, size);
+                let size_delta = size as i32 - removed_size as i32;
+                self.history_push(size_delta);
+            }
+            None => error!(
+                "heap_profiler: can't reallocate, pointer {} doesn't exist",
+                old_ptr
+            ),
+        };
+    }
+
+    fn free_profiler(&mut self, ptr: Ptr) {
+        debug!("heap_profiler: free({})", ptr);
+        if ptr != 0 {
+            match self.memory.remove(&ptr) {
+                Some(size) => {
+                    self.history_push(-(size as i32));
+                }
+                None => error!("heap_profiler: can't free, pointer {} doesn't exist", ptr),
+            };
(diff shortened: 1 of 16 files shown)


src/api/heap_profiler.rs
```diff
@@ -0,0 +1,142 @@
...
+            self.heap_history.pop_front().unwrap();
+        }
+        // TODO: trap/log error if elapsed failed
+        // very unlikely to happen!
+        self.heap_history
+            .push_back((change, self.started.elapsed().unwrap()));
+    }
+}
+
+#[host_functions(namespace = "heap_profiler", sync = "mutex")]
```
akegalj, line 94: @withtypes  in this PR actually I don't need `Mutex` at all. For this use case `Rc<RefCell<_>>` would be enough. `Rc<RefCell<_>>` would just be a bit more efficient solution (which is not that important for profiler, especially as there are bigger gains elsewhere).

If there will be other use cases similar like mine we could consider having third option that would wrap state with `Rc<RefCell<_>>` (sync = "ref"). This is just a small optimization that doesn't add much in this case, but it might be useful later

src/module/normalisation/heap_profiler.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    // modify wrapper function args
+    module.funcs.get_mut(fn_id).kind.unwrap_local_mut().args = local_vars;
+    Ok(())
+}
+
+// TODO: move to normalisation/utils.rs ?
+// Create a new local function that will have same signiture and same
+// instructions as the supplied local function.
+fn clone_function(module: &mut Module, fn_id: FunctionId, name: Option<String>) -> FunctionId {
```
akegalj, line 141: Cloning walrus local function is not that straightforward. This implements cloning of a function with all its parameters and instructions and can be reused elsewhere. If this is the case we could move these functions to `utils` module

tests/normalisation_patching_test/simple_heap_profiler.wat
```diff
@@ -0,0 +1,182 @@
...
+    (type (;4;) (func (param i32 i32) (result i32)))
+    (type (;5;) (func (param i32 i32 i32)))
+    (import "lunatic" "yield_" (func (;0;) (type 0)))
+    (import "heap_profiler" "malloc_profiler" (func (;1;) (type 3)))
+    (import "heap_profiler" "aligned_alloc_profiler" (func (;2;) (type 5)))
+    (import "heap_profiler" "calloc_profiler" (func (;3;) (type 5)))
+    (import "heap_profiler" "realloc_profiler" (func (;4;) (type 5)))
+    (import "heap_profiler" "free_profiler" (func (;5;) (type 1)))
+    (func $malloc_wrap (type 2) (param i32) (result i32)
+        block  ;; Reduction counter logic
```
akegalj, line 41: this test is unnecessary long due to reduction logic. We could change the way we test normalisation to shorten some of these (and future) examples

src/main.rs
```diff
@@ -21,20 +21,28 @@ struct Opts {
     /// All other arguments are forwarded to the .wasm file
     #[clap(min_values(0))]
     _args: Vec<String>,
+    /// Save heap profile to heap.dat
+    #[clap(short, long)]
+    profile: bool,
+    /// Output patched/normalised wasm to normalised.wasm
+    #[clap(short, long)]
+    normalised_out: bool,
```
akegalj, line 29: this flag was useful when debugging wasm so I kept it in PR

src/module/normalisation/heap_profiler.rs
```diff
@@ -0,0 +1,219 @@
...
+/// Profiling functions are imported from heap_profiler module.
+///
+/// Functions that can't be found in a module are ignored and error is logged.
+pub fn patch(module: &mut Module) {
+    // NOTE rusts global allocator __rust_alloc sometimes will
+    // execute aligned_alloc instead of malloc
+    ["malloc", "aligned_alloc", "calloc", "realloc", "free"]
+        .iter()
+        .for_each(|name| {
+            add_profiler_to(module, name).unwrap_or_else(|e| error!("{}", e));
```
akegalj, line 28: currently normalisation patch will log error instead of fail not to break tests. Better solution would be to let patch fail and refactor normalisation tests instead.

As profiler is not critical part of the system atm, simpler solution was chosen

src/api/process/process.rs
```diff
@@ -39,11 +43,11 @@ pub enum MemoryChoice {
 
 /// A lunatic process represents an actor.
 pub struct Process {
-    task: Task<Result<()>>,
+    task: Task<Result<HeapProfilerState>>,
```
withtypes, line 46: We can leave it for now but we will revert to `()` as host functions state will persist after a process halts.

src/api/heap_profiler.rs
```diff
@@ -0,0 +1,142 @@
...
+        if self.heap_history.len() > HISTORY_CAPACITY {
+            self.heap_history
+                .drain(0..(self.heap_history.len() - HISTORY_CAPACITY));
+        }
+    }
+
+    // Deallocate remaining memory. This is invoked when process exits
+    pub fn free_all(&mut self) {
+        // implicitly deallocate all left over child memory
+        self.memory.clone().iter().for_each(|(k, _)| {
```
withtypes, line 53: `free_all` is simple (sum all values in the map and clear the map). You don't need to reuse `free_profiler` (and therefore no need to clone the map), just implement it directly. Something like this:
```rust
        self.history_push(-(self.memory.values().sum::<u32>() as i32));
        self.memory.clear();
```
Do I miss something?

src/module/normalisation/heap_profiler.rs
withtypes: Can we contribute something to Walrus?

withtypes (changes requested):

src/module/normalisation/heap_profiler.rs
akegalj: I will open an issue there

src/api/heap_profiler.rs
akegalj: Yes, I think that should avoid the clone. ty

I wasn't aware of `sum` :| , and didn't want to reimplement `free_profiler` again so I decided to clone and reuse `free_profiler`for readability.

src/api/process/process.rs
akegalj: I will do another small PR where I will remove it and where I will remove merging profilers as it is done in the current form (so that processes can be observed as suggested by you here https://github.com/lunatic-solutions/lunatic/pull/44#discussion_r626264324)

src/api/heap_profiler.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    // Merge child process profile results into parent process.
+    //
+    // Usually this should be called when child process exits.
+    pub fn merge(&mut self, profiler: HeapProfilerState) {
+        self.started = std::cmp::min(self.started, profiler.started);
+        let started_delta = profiler.started.duration_since(self.started).unwrap();
+        // TODO sorted state merging of two sorted lists can be done in O(N+M)
```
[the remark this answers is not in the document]
akegalj: This should be resolved with https://github.com/lunatic-solutions/lunatic/pull/44/commits/5474f7ccb8749cc51df30f1f1b330ee8076fe2d7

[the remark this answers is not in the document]
akegalj: I hope location for test method is ok ? (within implementation module) Or should I refactor the test outside to seperate module?

akegalj:
It should be ready for another review.

I will open an issue on walrus repo to push some functionality upstream (https://github.com/lunatic-solutions/lunatic/pull/44#discussion_r626277687)
I will make another small PR refactor which resolves https://github.com/lunatic-solutions/lunatic/pull/44#discussion_r626269812 and https://github.com/lunatic-solutions/lunatic/pull/44#discussion_r626264324

src/api/heap_profiler.rs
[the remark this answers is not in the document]
withtypes: @akegalj yes, it's ok (I personally prefer it that way). You have "complied" :laughing:

PR #46, opened 2021-05-11
Enable heap profile observing while process is running

akegalj:
This PR addresses:
 * revert process return type to () as host functions state will persist after a process halts 
     - https://github.com/lunatic-solutions/lunatic/pull/44#discussion_r626269812 
 * have observable state/profiler as the process is running
     - https://github.com/lunatic-solutions/lunatic/pull/44#discussion_r626264324
 
 `HeapProfilerState` now contains all process profile histories observable as the process is running. Merging of all process profile histories is done after master process is finished (as oposed to after every child process finishes which was behaviour before this PR) just before it is written to file system (in src/main.rs).
 
This implementation will keep profile history of every child process in memory until master process finishes. For long running applications which will need profiling we might need to add aditional heuristic to drop some of the history. For example, we might want to drop histories of processes of all processes that are finished, or we might want to drop some of older finished process histories. At the moment this is not the case and we keep all process histories (note that every process history has its ringbuffer)

### DIFF
--- a/src/api/heap_profiler.rs
+++ b/src/api/heap_profiler.rs
@@ -221,22 +221,26 @@ fn merge_profiles() {
         memory: random_memory(),
         started: std::time::UNIX_EPOCH + Duration::from_millis(rand::random()),
         heap_history: random_history(),
+        processes: vec![],
     };
     let child = HeapProfilerState {
         memory: random_memory(),
         // we assume child is started after parent process
         started: parent.started + Duration::from_millis(rand::random()),
         heap_history: random_history(),
+        processes: vec![],
     };
     let mut parent_clone = HeapProfilerState {
         memory: parent.memory.clone(),
         started: parent.started.clone(),
         heap_history: parent.heap_history.clone(),
+        processes: vec![],
     };
     let child_clone = HeapProfilerState {
         memory: child.memory.clone(),
         started: child.started.clone(),
         heap_history: child.heap_history.clone(),
+        processes: vec![],
     };
     fn merge_simple(parent: &mut HeapProfilerState, child: HeapProfilerState) {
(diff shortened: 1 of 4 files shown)


src/api/heap_profiler.rs
```diff
@@ -23,20 +25,43 @@ impl HeapProfilerState {
...
-    // Merge child process profile results into parent process profile.
+    pub fn add_process(&mut self, process: <Self as HostFunctions>::Wrap) {
+        self.processes.push(process);
+    }
+
+    // Collect all process profiles into parent profile and remove them from process list
     //
-    // Usually this should be called when child process exits.
+    // Safety: this is safe to call only when all processes are finished.
+    pub fn collect_data(&mut self) -> anyhow::Result<()> {
```
akegalj, line 39: This is only one strategy, used for merging and writing merged profile to fs. We could implement other ways to use and show heap profile later

src/api/process/process.rs
```diff
@@ -133,31 +132,24 @@ impl Process {
         module: LunaticModule,
         function: FunctionLookup,
         memory: MemoryChoice,
-    ) -> Result<HeapProfilerState> {
+        profiler: <HeapProfilerState as HostFunctions>::Wrap,
+    ) -> Result<()> {
         let api = crate::api::default::DefaultApi::new(context_receiver, module.clone());
-        let ((profiler, _), fut) = Process::create_with_api(module, function, memory, api)?;
+        let ((p, _), fut) = Process::create_with_api(module, function, memory, api)?;
+        profiler.lock().unwrap().add_process(p.clone());
```
akegalj, line 139: we add process to profiler state when process is created

PR #47, opened 2021-05-28
Return an error code for failed TCP operations.

teymour-aldridge:


bkolobara:
Thanks for the contribution! Just for some context:

The original idea around error handling here was to return 2 integers back to wasm. The first would indicate:
* 0 - success
* 1 - error

The second would be the resource ID on success **or error id on error**. We can also communicate the error directly through the first argument with an integer >=1, as you did in this PR. And always return 0 for the second argument on error.

In this case I would also modify all the `*Result` types from `Err(String)` to `Err(())`. We never really use this error string and only convert it to 0 in the end. Could you do this too?

bkolobara:
After merging  the PR I realised it's not completely done :( I have pushed a few small cleanup fixes to main.

teymour-aldridge:
I'll submit a patch to `rust-lib` as well to update it to take into account the new changes.

PR #49, opened 2021-06-11
Add .wasm loading API for guests

bkolobara:
This PR adds support for loading .wasm files directly from guest code and spawning processes from them.

To get a better feeling for the Rust API check out these tests:
https://github.com/lunatic-solutions/rust-lib/blob/main/tests/module_loading_test.rs

It also allows you to define modules that will be used as imports for other modules.
These imports can shadow the VM exported functions and behave as a proxy to VM exports.

At this moment the "import" modules are somewhat limited, as they don't have access to the child process memory. And the VM host functions will always reference the child processes memory and not the proxy one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.15.1"
+version = "0.15.2"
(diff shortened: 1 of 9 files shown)


withtypes (approved):

bkolobara:
This was added as part of the rewrite.

PR #51, opened 2021-06-26
Fix a series of Clippy lints.

teymour-aldridge:


bkolobara:
Thanks @teymour-aldridge for running Clippy and fixing all issues! 🤗

PR #52, opened 2021-06-26
Add a basic CI workflow.

teymour-aldridge:


bkolobara:
This was added as part of the re-write.

PR #65, opened 2021-09-24
Udate wasmtime resolving breaking change

olanod:
Was trying to install lunatic and it was complaining about not being able to pick the right wasmtime version so updated it and resolved the breaking change that the new version has which doesn't seem to have any impact.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,9 +317,9 @@ checksum = "631ae5198c9be5e753e5cc215e1bd73c2b466a3565173db433f52bb9d3e66dba"
 
 [[package]]
 name = "cap-fs-ext"
-version = "0.17.0"
+version = "0.19.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a92bb3f74a5c7bab38a4670171299803994c100da721a71542acc0481f49aaab"
+checksum = "1bf5c3b436b94a1adac74032ff35d8aa5bae6ec2a7900e76432c9ae8dac4d673"
 dependencies = [
  "cap-primitives",
  "cap-std",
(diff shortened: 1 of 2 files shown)


olanod:
Btw, completely off-topic question that I hope you won't mind. I got here hopping from a wasmer repo discussion to async-wormwhole and then was glad to see a project that has done pretty much what I was trying to accomplish but was surprised to see that wasmer, which I intend to use due to the extra option of the llvm backend, was removed 😮 any particular reason for that removal?

bkolobara:
Thanks for the contribution! I think that the bigger issue here is the usage of the `git` "version". This breaks the build every time there is a [breaking commit](https://github.com/bytecodealliance/wasmtime/commit/bcf3544924880edc085e74497b0d3932909a456e) to wasmtime.

Previously we used the git version to get stuff that now landed in `0.30` and we can just use `0.30` from now on to avoid future breaking changes. If you could change the dependency to something like:
```
wasmtime = "0.30"
wasmtime-wasi = "0.30"
```
it would solve this kind of issues.

bkolobara:
> Btw, completely off-topic question that I hope you won't mind. I got here hopping from a wasmer repo discussion to async-wormwhole and then was glad to see a project that has done pretty much what I was trying to accomplish but was surprised to see that wasmer, which I intend to use due to the extra option of the llvm backend, was removed 😮 any particular reason for that removal?

`async-wormhole` was a pretty good proof of concept, but had some issues:
* It required rust nightly and would often break from one nightly version to another. I wouldn't even notice the breakage because I didn't update my rust version daily, but someone just starting out and trying to compile lunatic would be faced with a scary looking compile error. As you can see, it's not the nicest way to be introduced to a project :);
* The inline assembly would provide high performance. However, sometimes it would break backtraces on some architectures (Linux x64), but would work fine on Mac, Windows or Linux ARM. It was a bit of a nightmare to debug and I'm not sure the performance gains were worth the maintenance cost. If I wrote it again I would sacrifice this few `ns` for more stability;
* It didn't play well with libraries that use thread local storage and both, wasmtime and wasmer do use it. This forced us to maintain [a fork](https://github.com/lunatic-solutions/wasmer) of both libraries exposing some internals just to work around the TLS issues. Keeping the fork in sync with such active repositories was a full time job and any subtle change to the internals would silently break lunatic. Of course, you would only get a segfault that is impossible to reproduce because the scheduler moved your process in-between threads in an order that was unexpected. I spent days chasing this errors.

In the end wasmtime provided an `async` implementation that is similar to the way `async-wormhole` works (a separate stack), but is much better tested, works on stable rust and allows us to get rid of the forks we were maintaining. This significantly simplified my life and allowed me just to focus on the features that are important to lunatic.

Now that I have worked on this `async` rust + wasm stack for the last 2 years, I believe that there are alternative approaches to what async-wormhole and wasmtime are doing, but they would require the JIT compiler to be "async-aware". I would love to see Wasmer take this approach if they ever plan to release async host function support.

olanod:
The version in crates.io still doesn't have the "breaking change" so it'll be something to keep in mind for the next update ;)

PR #68, opened 2021-10-20
Feature/reply

withtypes:
Add simple support for message id which are local to the running process and reply ids which are set by Lunatic (and not guest) based on message ids. The idea is that the sender can match against reply id and a sent message (sending now returns message id).


withtypes:
I have added multiple changes:

- `Message` is not simpler, it's not an enum (`DataMessage`/`Signal`) but a struct with an attribute `is_signal`. The only question is which message id should a signal have as they don't have one before converting. It's much simpler and it simplifies host functions.
- separated the "scratch" area into "draft" and "reading" (maybe there is a better name)
- both are now non-option and this simplifies host functions, there are less traps, etc.:
  - there is always a draft. When you send a message, the draft gets replaced by a new one with the next message id
  - there is always a message being read. The first message is "a message to yourself" which you don't need to receive, it's already in the reading area. We could use it for something but now it's empty. For example, for getting initial data into the process.
 - `HashMapId` now starts from 1 but it doesn't use `NonZeroU64` as it makes usage complicated. For now it's just a convention not encoded in the type system. Having IDs start from 1 is useful for using 0 in host functions to denote `None`.
 - because we sometimes need to take ownership of message data (buffer, byte vector), this is now `Option`
 
In a separate PR I have removed tags completely as the use-case for waiting on reply is now covered. Because tags can be completely manipulated by guests, I see no advantage. If we want to select messages based on content, we can match against data buffer (first n bytes or an arbitrary slice).

A few ideas I didn't try out:

- a forward host function which would take the reading message and send it as is (same data, the source process and message id). Replies would come to the original process.
- a move data host function which would take data from reading and put it into draft. This way we can forward data but use the current message id and process id. Replies would come to this process.

bkolobara:
Great work! 🔥

Everything looks good to me. Most previous use cases should be supported, except:

- Waiting on multiple replies. The separate PR (#70) removes the waiting on a vector and can only match one reply at a time.
- Matching based on guest content.

I did actually have an implementation that matched messages based on the content of the buffer, similar to what Erlang does, but I decided to go with tags instead. The main reason I decided against it is that our buffers don't really have any structure and can't have an elegant way of matching content like Erlang. I wanted to let the guest decide the serialisation format, but if you decide to match on the first n bytes you are basically creating a new serialisation format that prefixes your serialisation with some data.

An arbitrary slice is even harder to work with, as most serialisers don't have strong guarantees about the layout of the serialisation. I doubt that it would be useful to provide such functionality.

In the end I just decided to remove the first n bytes and put them instead in the tag field, getting most of the benefits, but making the guest code much simpler. If you don't care about the matching you don't need to touch it. We can put it back into the buffer, but all the guests need to be aware that the first n bytes have a special meaning and should be set to 0 in most cases.

Is there maybe an even better way we could do this? So that the guest can express more complicated logic for the match. Something like _match all even "tags"_. Or match based on the structure and content of the serialised data. I couldn't come up with a good way of achieving this without forcing a specific serialisation format onto the guest.

One far fetched idea that came to my mind and an extreme case would be to move the message content from a buffer to something like [type imports](https://github.com/WebAssembly/proposal-type-imports/blob/master/proposals/type-imports/Overview.md) and structs from the [GC proposal](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md). This would result in a well defines serialisation format, allow us to match on partial or complete messages (like Erlang). It would work with minimal copy/serialisation overhead once more languages (with GC) get wasm support. For Rust and lower level languages we could provide a special serialisers that knows how to serialise into/from these structs. I never looked deeper into it as most of the proposals are still unsupported by Wasmtime.

withtypes:
@bkolobara regarding serialization, I think most use-cases would have a single byte as a tag (message type or something) and then the rest some guest-specific serialization. I'm more concerned about where "select" logic (which messages we want to receive/are waiting for) should go.

jtenner:
Hey what's the status on this? I was told I should work off of this branch for making bindings, and I just want to make sure this is going to be what we go with.

withtypes:
@jtenner sorry, I was distracted with some changes in the Rust lib/bindings and with the idea to move "selection" of messages into guests (still experimenting with this one, I will ask for some feedback) by using message handles (instead of a single-item reading buffer). This would give more freedom to the guest implementation which might be beneficial for AssemblyScript and maybe enable more idiomatic libs for guests. But I need a bit more time for this.

Regarding this PR, I think we definitely need reply ids as they are controlled by the runtime so guests cannot "fake" replies. And because tags were used for this mostly, I think they are redundant.

PR #74, opened 2021-12-26
Add version API.

teymour-aldridge:
This provides a way to obtain the version of the runtime. This is useful because if the client and host are of different versions, then this could cause unsafe behavior.

- `lunatic::version::major` - returns an integer containing the major version of the runtime
- `lunatic::version::minor` - returns an integer containing the minor version of the runtime
- `lunatic::version::patch` - returns an integer containing the patch version of the runtime

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 /target
 .DS_Store
+
+.vscode
+.idea
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -5,6 +5,7 @@ mod mailbox;
 mod networking;
 pub(crate) mod plugin;
 mod process;
+mod version;
 mod wasi;
 
 use std::future::Future;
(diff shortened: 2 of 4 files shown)


bkolobara:
Thanks @teymour-aldridge for taking onto this task!

Do we need additional granularity here? Like support for pre-release strings (e.g. `1.0.0-alpha.3`)? It may be useful to allow us testing pre-releases in lock step with guest versions and failing in scenarios where a guest for example expects `1.0.0-alpha.13` behaviour instead of `1.0.0-alpha.12`. In this case it would make sense to return one string altogether, but returning strings is not that easy without involving additional wasm proposals ([interface type canonical ABI](https://github.com/WebAssembly/interface-types/blob/40f157ad429772c2b6a8b66ce7b4df01e83ae76d/proposals/interface-types/CanonicalABI.md)).

It can also be that I'm over-optimizing unlikely scenarios.

teymour-aldridge:
Maybe we should have a flag for pre-release versions (e.g. a boolean) because standard semantic versioning does not hold for them?

bkolobara:
I will merge this as is. If we ever need the pre-release flag, we will have a better idea of what exactly we need and can add it then.

jtenner:
I'm a huge fan of this PR! Thank you very much.

PR #82, opened 2022-01-17
Implement udp functions

jtenner:
This pull request is an attempt to add support for UdpSocket to lunatic. The work I suspect needs to be done:

- [x] `udp_bind` to address with exact same signature as `tcp_bind`
- [x] `udp_connect`
- [x] `drop_udp_socket`
- [x] `clone_udp_socket`
- [x] `udp_send_to`
- [x] `udp_read`
- [x] `set_udp_socket_broadcast`
- [x] `get_udp_socket_broadcast`
- [x] `set_udp_socket_ttl`
- [x] `get_udp_socket_ttl`

and in the mailbox namespace

- [x] `push_udp_socket`
- [x] `take_udp_socket`

Looks like `udp_bind` wasn't so bad.

EDIT:

Rename the udp_read function to udp_receive, to stay consistent in the naming.
- [x] Add the udp_receive_from function, currently you can udp_send_to without connecting, but can't receive.
- [x]  Add udp_send, same reason as the previous point. You should be able to send when connected.
- [ ]  Add some documentation that udp_send & udp_receive will fail if you didn't call udp_connect before.
- [ ]  Add all the new host functions to this test file: https://github.com/lunatic-solutions/lunatic/blob/main/wat/all_imports.wat


debug.log
```diff
@@ -0,0 +1 @@
+[0118/133135.552:ERROR:registration_protocol_win.cc(102)] CreateFile: The system cannot find the file specified. (0x2)
```
bkolobara, line 1: I don't think we need this file checked in :)

jtenner: Oh. Which file?

bkolobara: The one that I commented on, `debug.log`.

bkolobara:
Thanks for contributing this @jtenner!

It looks pretty good. I would just propose a few changes:

- [ ] Rename the `udp_read` function to `udp_receive`, to stay consistent in the naming.
- [ ] Add the `udp_receive_from` function, currently you can `udp_send_to` without connecting, but can't receive.
- [ ] Add `udp_send`, same reason as the previous point. You should be able to send when connected.
- [ ] Add some documentation that `udp_send` & `udp_receive` will fail if you didn't call `udp_connect` before.
- [ ] Add all the new host functions to this test file: https://github.com/lunatic-solutions/lunatic/blob/main/wat/all_imports.wat

src/api/mailbox.rs
```diff
@@ -592,3 +608,56 @@ fn receive(
         }
     })
 }
+
+//% lunatic::message::push_tcp_stream(stream_id: u64) -> u64
```
ejc123, line 612: Change

ejc123, line 612: Change tcp to udp.  There are a few other places too.

jtenner: Oh wow. I didn't see this before. Thanks for the heads up!
👍

debug.log
jtenner: debug.log. Got it. Why was I so confused about this? lol

jtenner:
Apparently I had receive_form implemented under the name "receive" by accident. I switched the implementation to receive_from and then implemented receive.

src/api/networking.rs
```diff
@@ -854,3 +1015,564 @@ fn socket_address(
...
+
+//% lunatic::networking::udp_receive_from(
+//%     socket_id: u64,
+//%     buffer_ptr: u32,
+//%     buffer_len: u32,
+//%     addr_type: u32,
+//%     addr_u8_ptr: u32,
+//%     port: u32,
+//%     flow_info: u32,
+//%     scope_id: u32,
```
bkolobara, line 1172: This fields are c/p from another function, but don't exist in this signature.

src/api/networking.rs
```diff
@@ -854,3 +1015,562 @@ fn socket_address(
...
+//% * If broadcast traps.
+fn get_udp_socket_broadcast(caller: Caller<ProcessState>, udp_socket_id: u64) -> Result<i32, Trap> {
+    let socket = caller
+        .data()
+        .resources
+        .udp_sockets
+        .get(udp_socket_id)
+        .or_trap("lunatic::networking::get_udp_socket_broadcast")?;
+
+    let result = socket
```
jtenner, line 1386: This looks inconsistent. The result local should be on line 1370

src/api/networking.rs
```diff
@@ -854,3 +1023,615 @@ fn socket_address(
...
+fn udp_local_addr(
+    mut caller: Caller<ProcessState>,
+    udp_socket_id: u64,
+    id_u64_ptr: u32,
+) -> Result<u32, Trap> {
+    let udp_socket = caller
+        .data()
+        .resources
+        .udp_sockets
+        .get(tcp_listener_id)
```
bkolobara, line 1613: The compilation fails on this line with:
```
Compiling lunatic-runtime v0.7.5 (/home/runner/work/lunatic/lunatic)
error[E0425]: cannot find value `tcp_listener_id` in this scope
    --> src/api/networking.rs:1613:14
     |
1613 |         .get(tcp_listener_id)
     |              ^^^^^^^^^^^^^^^ not found in this scope
```

src/message.rs
```diff
@@ -110,6 +116,26 @@ impl DataMessage {
         None
     }
 
+    /// Takes a TCP stream from the message, but preserves the indexes of all others.
```
bkolobara, line 119: This docs also need a fix.

src/api/networking.rs
```diff
@@ -854,3 +1023,615 @@ fn socket_address(
...
+            )
+            .or_trap("lunatic::networking::udp_bind")?;
+
+        Ok(result)
+    })
+}
+
+//% lunatic::networking::drop_udp_listener(tcp_listener_id: i64)
+//%
+//% Drops the UCP listener resource.
```
bkolobara, line 1095: Typo

src/api/networking.rs
```diff
@@ -854,3 +1023,615 @@ fn socket_address(
...
+//% * 1 on error      - The error ID is written to **opaque_ptr**
+//% * 9027 on timeout - The socket receive timed out.
+//%
+//% Reads data from the connected udp socket and writes it to the given buffer.
+//%
+//% Traps:
+//% * If the stream ID doesn't exist.
+//% * If **buffer_ptr + buffer_len** is outside the memory.
+//% * If **i64_opaque_ptr** is outside the memory.
+//% * If the socket hasn't been connected yet. (use socket.connect() to connect to an address)
```
bkolobara, line 1128: I don't think this function will trap if the socket is not connected yet. It will just just return an error. On line `1158` it captures the error and writes it the the return value.

src/api/networking.rs
```diff
@@ -854,3 +1023,615 @@ fn socket_address(
...
+            addr_type,
+            addr_u8_ptr,
+            port,
+            flow_info,
+            scope_id,
+        )?;
+
+        if let Some(result) = tokio::select! {
+            _ = async_std::task::sleep(Duration::from_millis(timeout as u64)), if timeout != 0 => None,
+            result = UdpSocket::bind("127.0.0.1:0") => Some(result)
```
bkolobara, line 1298: I believe the `connect` function is meant to be used on an existing `UdpSocket`, and we shouldn't create a new on `connect`.

It sets an existing `UdpSocket` to filter out any packets coming from a certain source. E.g. I bind to `localhost:4000`, get a socket back, connect this socket to `x.x.x.x:3000`. Now I listen to all incoming messages on `localhost:4000` that are comming from `x.x.x.x:3000`

If you assign every time a random address on `connect` you lose the possibility to control your local address.

src/api/networking.rs
```diff
@@ -854,3 +1023,615 @@ fn socket_address(
...
+                .or_trap("lunatic::networking::udp_connect")?;
+            Ok(result)
+        } else {
+            // Call timed out
+            Ok(9027)
+        }
+    })
+}
+
+//% lunatic::networking::clone_tcp_stream(udp_socket_id: u64) -> u64
```
bkolobara, line 1330: This should say clone_~tcp~upd_stream

src/api/networking.rs
```diff
@@ -854,3 +1023,615 @@ fn socket_address(
...
+//% * 9027 on timeout - The socket send timed out.
+//%
+//% Gathers the buffer and writes it to the connected socket, and returns the number of bytes written.
+//%
+//% Traps:
+//% * If the stream ID doesn't exist.
+//% * If **buffer_ptr + (buffer_len)** is outside the memory.
+//% * If **i64_opaque_ptr** is outside the memory.
+//% * If the socket hasn't been connected yet. (use socket.connect() to connect to an address)
+//% * If send() traps.
```
bkolobara, line 1544: In both cases an error is returned and there is not trap.

src/api/networking.rs
jtenner: Thanks for catching this. I misunderstood how it was supposed to work, and just followed the examples I found on the UdpSocket docs.

jtenner:
```
failures:

---- api::tests::import_filter_signature_matches stdout ----
thread 'api::tests::import_filter_signature_matches' panicked at 'called `Result::unwrap()` on an `Err` value: unknown import: `lunatic::networking::local_addr` has not been defined', src/api/mod.rs:89:55
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Interesting. Looks like I missed something here. What is this?

src/api/networking.rs
```diff
@@ -854,3 +1024,605 @@ fn socket_address(
...
+//% * 1 on error      - The error ID is written to **opaque_ptr**
+//% * 9027 on timeout - The socket receive timed out.
+//% Reads data from TCP stream and writes it to the buffer.
+//%
+//% Traps:
+//% * If the stream ID doesn't exist.
+//% * If **buffer_ptr + buffer_len** is outside the memory.
+//% * If **i64_opaque_ptr** is outside the memory.
+//% * If **i64_dns_iter_ptr** is outside the memory.
+fn udp_receive_from(
```
jtenner, line 1194: It looks like this function isn't correct. I will work on this tomorrow.

bkolobara:
Thank you @jtenner for the contribution. I did some small changes to it (docs and tests). Could you allow me to push these changes to the PR? There is a checkbox on this page on the right side saying something like `Allow edits from maintainers`.

Everything else looks good and i think it can be merged. 🥳

PR #83, opened 2022-01-19
Return 2 on network timeouts instead of an error.

### LINKED ISSUE #80: tcp_read return values inconsistent
https://github.com/lunatic-solutions/lunatic/blob/944935e857e53dc0278cbb4f16bb8a50f097d460/src/api/networking.rs#L709

Per @bkolobara 's request, I have opened this issue to document a small inconsistency in the way socket reads occur.

Per our discussion, it should probably return

```
Ok -> 0
Err -> 1
Ok with no bytes read (socket closed normally) -> 0
Timeout -> 9027
```

Thanks!

teymour-aldridge:
Fixes #80

### DIFF
--- a/src/api/networking.rs
+++ b/src/api/networking.rs
@@ -193,6 +193,7 @@ pub(crate) fn register(
 //% Returns:
 //% * 0 on success - The ID of the newly created DNS iterator is written to **id_u64_ptr**
 //% * 1 on error   - The error ID is written to **id_u64_ptr**
+//% * 9027 if the operation timed out
 //%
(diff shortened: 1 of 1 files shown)


src/api/networking.rs
```diff
@@ -193,6 +193,7 @@ pub(crate) fn register(
 //% Returns:
 //% * 0 on success - The ID of the newly created DNS iterator is written to **id_u64_ptr**
 //% * 1 on error   - The error ID is written to **id_u64_ptr**
+//% * 2 on timeout
```
bkolobara, line 196: I think it should be `9027` instead of `2` here.

PR #92, opened 2022-01-27
Implement first class testing

jtenner:
Since I'm new to this project, I wasn't sure exactly how to implement the `"_lunatic_test*"` pattern, so I naively just collected all the join handles asynchronously, and awaited them. This compiles.

Any thoughts on this?


bkolobara:
With https://github.com/lunatic-solutions/lunatic/pull/101 this is already implemented.

PR #93, opened 2022-02-02
Tap testing - initial support

jtenner:
This pull request aims to add TAP testing support via a couple of mechanisms. 

- [x] Test functions are automatically called:
  - [x] Primary test entry is an exported `__lunatic_start_test` method
  - [x] Any function exported with the name `__lunatic_test` at the start will automatically be called and spawned in parallel.
- [x] A couple of exposed host functions
  - [x] `lunatic::test::add_test(parent_id: u64, name: u32, name_len: u32, negated: u32)` (Add a test to the test suite, and denote if it *should* fail)
  - [x] `lunatic::test::add_test_comment(node_id: u64, name: u32, name_len: u32)` (Add a comment to a test)
  - [x] `lunatic::test::test_ok(test_id: u64)`
- [ ] Write a file to the filesystem with tap output `--test ./build/output.tap`
- [ ] Consume the tap output in memory if `--report` flag is passed

### DIFF
--- a/src/api/process.rs
+++ b/src/api/process.rs
@@ -295,7 +295,6 @@ pub(crate) fn register(
 //% Create a new configuration for an environment.
 fn create_config(mut caller: Caller<ProcessState>, max_memory: u64, max_fuel: u64) -> u64 {
     let max_fuel = if max_fuel != 0 { Some(max_fuel) } else { None };
-    caller.data_mut();
     let config = EnvConfig::new(max_memory as usize, max_fuel);
     caller.data_mut().resources.configs.add(config)
 }
--- a/src/api/process.rs
+++ b/src/api/process.rs
@@ -295,6 +295,7 @@ pub(crate) fn register(
 //% Create a new configuration for an environment.
 fn create_config(mut caller: Caller<ProcessState>, max_memory: u64, max_fuel: u64) -> u64 {
     let max_fuel = if max_fuel != 0 { Some(max_fuel) } else { None };
+    caller.data_mut();
     let config = EnvConfig::new(max_memory as usize, max_fuel);
     caller.data_mut().resources.configs.add(config)
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1154,6 +1154,7 @@ dependencies = [
  "pretty_assertions",
  "semver",
  "serde",
+ "testanything",
  "tokio",
  "uuid",
  "wabt",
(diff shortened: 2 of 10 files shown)


src/api/process.rs
```diff
@@ -295,6 +295,7 @@ pub(crate) fn register(
 //% Create a new configuration for an environment.
 fn create_config(mut caller: Caller<ProcessState>, max_memory: u64, max_fuel: u64) -> u64 {
     let max_fuel = if max_fuel != 0 { Some(max_fuel) } else { None };
+    caller.data_mut();
```
jtenner, line 298: This looks unecessary. Whoops

jtenner:
Given that this testing implementation is rather bare bones, It's well to note that the TAP protocol version 13 supports YAML messages that follow directive comments. There are also other directive types that might be useful to compile to. 

https://testanything.org/tap-version-13-specification.html

Because tests can be nested in TAP13, I created the wasm linked functions to be structure aware. Currently in this implementation, it would cause *every* test to be flattened into a single group, simply to "make it work."

Future improvements might be:
- [ ] Implementing proper TAP output with a `Message` enum having attached messages to `TestNode`s
  - [ ] Comment(String)
  - [ ] Fail(ActualString, ExpectedString, CommentString)
  - [ ] Todo(String)
 - [ ] Test Skipping Directive
 - [ ] Todo Directive (which is different than a Todo mesage [see above])

bkolobara:
Thank you for experimenting with this @jtenner!

One of the goals of lunatic is also to be usable as a library. You should be able to extend the runtime and embed it into your rust application. To be able to do this we will probably need a way for users to provide their own `ProcessState`. I'm currently looking into how to do this in an ergonomic and performant way.

Maybe we can use this "bring your own process state" functionality to also provide a special process state that holds test data when running inside the test mode. In this case we could avoid the global TEST mutex and also put the testing implementation behind a feature flag and let embedders choose if they need it or not.

It would also be cool if we could stream the test results as they are done, one by one. In the current implementation you need to wait on all tests to finish before printing anything out. It happens often to me that during testing a process gets stuck and never finishes. Mostly because it is waiting on a message. This means that I would not be able to even identify which of the tests is stuck.

However, my biggest issues with this approach is that we require a TAP consumer to get any feedback from the tests. I have been asking a bit around and it doesn't seem like anyone is using TAP in their testing setups. I'm just afraid of locking ourself in a protocol that has barley any adoption (< 1% of all JS npm testing library downloads, ~ 0% in Rust), and that may not play well with some of the other testing scenarios that @teymour-aldridge mentioned in this comment: https://github.com/lunatic-solutions/rust-lib/issues/8#issuecomment-922344188

For example:
- property testing (e.g. [proptest](https://docs.rs/proptest))
- model checking
- [simulation](https://sled.rs/simulation.html)
- fuzzing
- chaos engineering

jtenner:
I have a bunch of comments on this approach. I couldn't agree more. I happen to like TAP but it doesn't seem too widely adopted as you have described. However, this doesn't mean we can't standardize on something *better.*

I'm currently in the process of working on something involving assemblyscript and a flash player emulator, but I promise to swing around to this again too after I investigate native_tls support which I feel is plenty more important for the time being.

Issue #69, opened 2021-10-20
New plugin system


bkolobara:
Lunatic already has a [plugin system](https://github.com/lunatic-solutions/lunatic/blob/main/src/plugin.rs) that is not well documented or particularly useful in the current state. It started out as a way to dynamically change host functions with WebAssembly modules, but ended up being a tool to modify WebAssembly modules before instantiating them.

I want to use this issue to discuss pros and cons of different approaches to a plugin system and figure out what the best path forward would be. In my grand vision plugins would be an easy way to extend the VM without needing to recompile it, just by adding Wasm modules to it. First of all this means extending the host functions. It could also be a playground for adding new functionality to the VM; we can test it as a plugin behind a feature flag and later provide a native implementation. This could potentially also improve development speed, because plugins can be dynamically loaded and added to environments inside one VM instance. Testing plugins or comparing performance between competing implementations becomes trivial, just spawn different environments, attach different plugins to them and run the same code. You could implement plugins in this case in any language compiling to Wasm, you would not only be limited to rust when extending the VM.

Another use case could be replacing existing functionality with custom implementations. One example of this would be to provide a custom networking layer for the distributed lunatic implementation.

We can go even further here and allow people to load native shared libraries that are exposed as host functions (an idea of [@jtenner](https://github.com/jtenner)). Or even allow plugins to modify the loaded Wasm code before JIT compiling it, an approach that our [heap profiler plugin](https://github.com/lunatic-solutions/lunatic/tree/main/plugins/heap_profiler) is taking.

The most important questions I would like answered here are:
- [ ] How do we provide a good user experience around plugins? If creating plugins is too complicated and the complexity around them is hard to manage, nobody is going to be using them.
- [ ] What kind of plugin system would give us the best trade-off between user experience and performance?
- [ ] Should we allow native code (that could potentially crash the whole VM) plugins?

## 1. Plugins that provide host functions

Originally I wanted lunatic to be just a set of core APIs (networking, filesystem, processes, ...) and every other functionality would be implemented as a plugin in WebAssembly on top of it. Around the same time Wasmtime got support for [module-linking](https://github.com/WebAssembly/module-linking/blob/main/proposals/module-linking/Explainer.md) and I saw a great opportunity to actually accomplish this. The idea was simple, you could add new host functions to processes by defining them inside WebAssembly plugins. These host functions would provide higher level abstractions on top of the core APIs or even shadow/change the core APIs.

[...]

PR #96, opened 2022-02-08
Remove plugin API & split project into crates

bkolobara:
This PR does 2 changes:

### 1. Removes the plugin subsystem + associated host function APIs

In the current shape the plugin system is undocumented and not particularly useful. It has a lot of code that touches many parts of the codebase. We don't even have a good use case for it at the moment and this PR removes is just to simplify the rest of our codebase.

The following issue tracks some ideas around a more useful plugin system: #69.

### 2. Splits some host functions into separate crate:

- [x] Error handling
- [x] Messaging
- [x] Networking
- [x] Version
- [ ] Process/Module/Environment
- [x] WASI

This change will allow us to feature gate some of the dependencies if lunatic is used in an embedded context and not all host functions are required. It is also the first step in abstracting the process state into something more generic that can be extended.

This required some other parts of the runtime to be separated out into crates too, to avoid circular dependencies. The host functions that deal with process spawning, modules and environments are not separated out. They are too entangled with the `Node` struct (distributed) and can't be pulled out at the moment (@withtypes).

WASI could also be separated out at one point.


sleipnir:
> This change will allow us to feature gate some of the dependencies if lunatic is used in an embedded context

This is likely to be particularly useful in integrating with Krustlet, if there is a way to run the lunatic runtime programmatically

bkolobara:
@sleipnir I'm very much interested in bringing lunatic support to Krustlet. It should be even possible at this stage to turn lunatic into a Krustlet provider. 

Lunatic has 2 features that other providers don't offer:
1. A concurrency model based on processes
2. Networking primitives

This should allow people to already deploy some useful applications, like web apps, to k8s. Until now I just saw simple demo apps running on Krustlet. I don't have much time to look into it at the moment, but could assist if someone wants to take onto the task. Looking at some of the other providers could be a good starting point, like https://github.com/krustlet/krustlet/pull/677.

sleipnir:
Yes I have been following and researching how this would be implemented for a long time, I hope to soon be able to provide a proper implementation for Lunatic

withtypes (approved):
I like the idea and the code seems ok. I've tested it with my toy distributed ping-pong example. I can try and move wasi to a separate crate...

src/message.rs
```diff
@@ -1,194 +1 @@
-/*!
```
withtypes, line 1: I noticed this wasn't deleted...

PR #105, opened 2022-06-04
Fix homebrew typo

pingiun:


bkolobara:
Thanks for the fix @pingiun!

PR #106, opened 2022-06-05
add send_after and cancel_timer host functions

zhamlin:


bkolobara:
That's perfect timing, I just need `send_after`. Looks great, thanks for the contribution @zhamlin.

Issue #95, opened 2022-02-03
Remove timeout support on networking read/writes


bkolobara:
Tcp stream's `read` & `write` and udp's `send` & `receive` functions take a `timeout` argument. If it's greater than 0, the function will only block for the given amount in _ms_ before returning a timeout code. I would like to remove this functionality.

As someone who has implemented some of these functions, I believe it's almost impossible to reason about code using them. All of the listed functions will return 1 or more bytes and it's hard to express behaviour like "wait max 5 sec on the next 1 Mb". You need to create a state machine and keep counting and reducing time yourself. The complexity of using them correctly is just too high. Has anyone ever successfully used them for anything?

It makes the implementation also way more complicated than it needs to be. We need to use a macro that fuses futures together, set a timer in the kernel on each `read/write` call, make sure all the futures are cancel safe, etc. It makes the implementation really hard to reason about without much of a benefit. I also believe that it would lead to a high performance hit if the timeout argument was used in practice, but I never benchmarked it.

The most common scenario I assume is waiting for some structured data (e.g. HTTP request) and giving up if it doesn't arrive on time. And this can't be easily expressed with the current timeout functionality.

I propose instead to handle such cases by spawning a process, collecting all the data from the TcpStream until a valid request is formed and send it back to the parent. The message `receive` function of the parent would still keep the timeout parameter. One of lunatic's goals is to make process spawning extremely performant, so that developers can use it without fear of a performance hit. I think this approach aligns better with lunatic's design in general.

PR #107, opened 2022-06-10
Switch to tokio from async-std

kosticmarin:
Migrate to tokio from async-std as proposed in #94.

As discussed in the issue above, `TcpStream` can be wrapped into an `Arc` but this is not enough since we use `tokio::select!` to handle timeouts which requires async api. To overcome this, we can simply wrap the `TcpStream` into `Arc` with a `Mutex`, this brings some locking overhead.

Also two tests in rust-lib fail against this build of the runtime, could use some help here to understand better what is happening.

```
test lunatic::mailbox::tests::mailbox_link - should panic ... FAILED
test lunatic::process::tests::spawn_link_test - should panic ... FAILED
```

Traces can be produced with `WASMTIME_BACKTRACE_DETAILS=1 RUST_LOG=lunatic=trace cargo test`

```
[2022-06-10T09:24:40Z WARN  lunatic_process] Process 2 failed, notifying: 1 links
[2022-06-10T09:24:40Z DEBUG lunatic_process] wasm trap: wasm `unreachable` instruction executed
    wasm backtrace:
        0: 0x82c2d - panic_abort::__rust_start_panic::abort::hca696c21df4c435e
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/panic_abort/src/lib.rs:84:17
                  - __rust_start_panic
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/panic_abort/src/lib.rs:38:5
        1: 0x82844 - rust_panic
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:746:9
        2: 0x827c1 - std::panicking::rust_panic_with_hook::h7811fbe84dbe0b81
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:716:5
        3: 0x2dd2d - std::panicking::begin_panic::{{closure}}::h2ae1f2e0bc0d2948
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:617:9
        4: 0x2dc70 - std::sys_common::backtrace::__rust_end_short_backtrace::h148ca43cd1d584f3
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/sys_common/backtrace.rs:138:18
        5: 0x2dddc - std::panicking::begin_panic::hfbb35725cbd3b0a6
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:616:12
        6: 0x2cc23 - lunatic::mailbox::tests::mailbox_link::{{closure}}::he61bd5cfcbf9a086
                        at /home/marin/github/lunatic-solutions/rust-lib/src/mailbox.rs:295:13
        7: 0x12f34 - core::ops::function::FnOnce::call_once::hcecd41f517596a32
                        at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:227:5
        8: 0x11690 - lunatic::mailbox::type_helper_wrapper::h67f4a5cb34e065da
                        at /home/marin/github/lunatic-solutions/rust-lib/src/mailbox.rs:269:5
        9: 0x1af1 - _lunatic_spawn_by_index
                        at /home/marin/github/lunatic-solutions/rust-lib/src/host/mod.rs:57:5
       10: 0x95550 - <unknown>!_lunatic_spawn_by_index.command_export
```


bkolobara:
Great work! I'm really looking forward to landing this one.

Just some random notes. Unfortunately, putting the `TcpStream` into a `Mutex` will not work. I did the same when I attempted to replace async_std with tokio. The issue is that writing and reading need to be simultaneous for the majority of use-cases and will lead to strange issues if they need to be exclusive.

Imagine the following scenario. You have two processes one reading from `TcpStream` and another writing to it. On the other side you just have an echo server. Now, the writing process locks, sends a message to the external server and releases the lock. The external server sends the same data back. Now the reading process locks and reads the data. So far so good. However, both processes are scheduled independently, this means that the reading process may attempt to read something before the writer. In this case the reader will block forever, because there is not data coming in. For data to be available, the writer would need to write to the echo server, but it can't acquire the lock while the reader is waiting. And you deadlocked yourself.

The good news is that we can remove the `tokio::select!` part, this timeouts are mostly useless: https://github.com/lunatic-solutions/lunatic/issues/95

I will have a deeper look into why the panic doesn't break the link correctly.

bkolobara:
It looks like `.await`ing a result of `JoinHandle` does different things in `tokio` and `async_std`. In `async_std` it will just return the result of the async task, but in `tokio` it will return `Result<result of async task>`, depending if the task panicked. That's why the tests are are not able to detect a panic. Tokio's `JoinHandle` basically wrapped the result into an `Ok(Err(...))`.

You will need to `.unwrap()` once more in:
1. https://github.com/lunatic-solutions/lunatic/blob/tokio/src/mode/execution.rs#L178
2. https://github.com/lunatic-solutions/lunatic/blob/tokio/src/mode/cargo_test.rs#L241

I think `.unwrap()` is ok here, as it would only be an issue if we have a panic inside a host function and this should "never" be the case. Maybe we could do some smarter error handling around this in the future.

bkolobara (approved):
Looks good to me :)

withtypes:
@kosticmarin can you just resolve these two conflicts and we can merge it then

PR #111, opened 2022-06-27
Distributed v2

### LINKED ISSUE #94: Switch to tokio.rs from async_std

withtypes:
New distributed implementation introduces the idea of control nodes. Control nodes keep information about lunatic nodes, Wasm modules, environments, and perform authentication. Lunatic nodes communicate directly after they register via control nodes.
We provide a simple implementation of a control node for now, but the idea is to have a distributed and persistent control nodes (backed by a distributed key-value database).

## Architecture

Code is kept in two crates: `lunatic-distributed` and `lunatic-distributed-api` (host functions for Wasm).

There are two changes in non-distributed lunatic:
 - addition of environments: they act as a sandbox for a group of processes. Newly spawned processes inherit the parent's environment.  Processes can only communicate inside the environment, and process ids are unique only inside the environment (and node).
 - processes are not resources anymore (kept in process state resources) and guests don't use a resource handle (opaque reference), but the actual process id (`u64`). Processes are identified by both node id (`u64`) and process id.
When a lunatic node starts, it registers to a control node by sending the listen address, and authentication data. On successful registration, it receives a node id. Connections between nodes are lazy and established only when one node needs to communicate with another.

Distributed API introduces new host functions:
 - `nodes_count` and `get_nodes` which currently return only a list of other node ids
 - `node_id` returns the current node id
 - `module_id` returns the global id assigned to a currently running Wasm module. This id is used when spawning on other nodes (they fetch modules from a control node). 
 - `spawn` is similar to local spawn but also accepts a node id. It doesn't support linking currently
 - `send` and `send_receive_skip_search` also receive 
additional node id parameter.
  
Lunatic can also run in a non-distributed way as it does now.

The diagram below shows how process state depends on other distributed lunatic structures.

## Command Line Arguments

Command lines arguments were kept as in a previous version of the distributed:
 - `--node` sets the address and port for the node-to-node communication
 - `--control` sets the address of the control node
 - `--control-server` runs the control server bound to `--control` address
 - `--no-entry` run the node without a Wasm module

Here is a small script that runs 4 nodes. The script assumes this repository is located in the `lunatic` directory and `lunatic-rs` contains the lunatic Rust library repository.

```bash
#!/bin/bash
set -e
SCRIPT=$(readlink -f "$0")
BASEDIR=$(dirname "$SCRIPT")

export RUST_LOG="info"

cd $BASEDIR/lunatic
cargo build --release

cd $BASEDIR/lunatic-rs
cargo build --release --examples

prefix() {
    # A helper to add prefix to each stdout/stderr line. It's useful to tell nodes apart
    local prefix="$1"
    shift
    "$@" > >(sed "s/^/$prefix: /") 2> >(sed "s/^/$prefix: /" >&2)
}

lunatic=$BASEDIR/lunatic/target/release/lunatic

prefix "A" $lunatic --control 127.0.0.1:10410 --node 127.0.0.1:10001 --control-server --no-entry &
prefix "B" $luantic --control 127.0.0.1:10410 --node 127.0.0.1:10002 --no-entry &
prefix "C" $lunatic --control 127.0.0.1:10410 --node 127.0.0.1:10003 --no-entry &
prefix "D" $lunatic --control 127.0.0.1:10410 --node 127.0.0.1:10004 "$BASEDIR/lunatic-rs/target/wasm32-wasi/release/examples/distributed.wasm"

sleep 2
killall lunatic
```

## Other Updates

Thanks @kosticmarin we have also replaced `async-std` with `tokio` which resolves #94.
He has also worked on replacing plain TCP with QUIC. This should also be implemented soon.

## Future Work

Here are some things for the future: 
 - currently distributed linking is not implemented and linking in general needs some refactoring
 - we are replacing plain TCP with QUIC which will give us streams and encryption
 - better errors in node-to-node communication
 - authentication
 - bandwith control with chunking and QUIC streams
 - node attributes (region, provider, etc.)
 - trusted, untrusted, and privelaged nodes
 - WASI for distributed, virtual FS


crates/lunatic-process/src/env.rs
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    pub fn get_process(&self, id: u64) -> Option<Arc<dyn Process>> {
+        self.processes.get(&id).map(|x| x.clone())
+    }
+
+    pub fn add_process(&self, id: u64, proc: Arc<dyn Process>) {
+        self.processes.insert(id, proc);
+    }
```
bkolobara, line 31: We also need a `remove_process` once it finishes running.

withtypes:
@bkolobara there were no conflicts, tests fail in lunatic-rs because of the incompatible 0.9.0 version. I've updated both lunatic and lunatic-rs projects.

I'm currently trying to make quic work (there is some problem) and also simplify TLS cert management for node-to-node communication. This is done in #114

bkolobara (changes requested):
Looks good! I think we should merge this already after a few small fixes, so it doesn't continue growing.

withtypes:
I have removed unnecessary traps on send/link.

crates/lunatic-process-api/src/lib.rs
```diff
@@ -633,21 +633,29 @@ fn link<T: ProcessState + ProcessCtx<T>>(
...
+            .0
+            .send(Signal::Link(tag, process))
+            .expect("The Link signal is sent to itself and the receiver must exist at this point");
+    } else {
+        // Link died (currently death reason normal is sent)
+        caller
+            .data_mut()
+            .signal_mailbox()
+            .0
+            .send(Signal::LinkDied(process_id, tag, DeathReason::Normal))
```
bkolobara, line 654: I think we should use another `DeathReason` here. Erlang uses the `noproc` exit signal, that will fail the process by default if signals are not trapped. We should maybe use something similar to indicate that the action of linking has failed, as we can't really tell why the process finished (normal, failure, ...).

Note: The Erlang behaviour is actually a bit complicated. If it's a local process (on the same node) and the check of process existence is "fast", the caller raises a `noproc` exception, otherwise it receives a `noproc` signal. This is done like this because of historical reasons. We don't need to keep this inconsistent behaviour and can always send a signal. Raising an exception in the guest wouldn't even be an option for us.

crates/lunatic-process-api/src/lib.rs
```diff
@@ -665,21 +673,19 @@ fn unlink<T: ProcessState + ProcessCtx<T>>(
...
+    if let Some(process) = process {
+        process.send(Signal::UnLink(Arc::new(this_process)));
+
+        // Send unlink signal to itself
+        caller
+            .data_mut()
+            .signal_mailbox()
+            .0
+            .send(Signal::UnLink(process))
+            .expect("The signal is sent to itself and the receiver must exist at this point");
```
bkolobara, line 687: The unlink signal should always be sent to itself, even if the other process doesn't exist. Keeping the link locally doesn't do anything and just consumes resources.

withtypes: OK, but that was the problem as `Signal::UnLink` containts `Process` which in this case doesn't exist.

withtypes: I will change `UnLink` to have process id, not `Process`. It's simple.

crates/lunatic-process-api/src/lib.rs
withtypes: I've added `NoProcess` reason but the behaviour is the same as with normal.

PR #114, opened 2022-07-01
Replace tcp with quic for distributed nodes

kosticmarin:
Use QUIC for node-to-node and node-to-control connections. On older Linux kernel, the `s2n-quic` crate didn't work (it wouldn't send more that ~2000 bytes) so now default is `quinn`. During registration, nodes generate a certificate signing request which is signed by a control node. To simplify testing, you can use the `--test-ca` command line flag which will simply use built-in certificate and keys (so this is only for testing!).

For example:
```bash
lunatic --control 127.0.0.1:10410 --control-server --no-entry --test-ca & # run the control server and node
sleep 1
lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10002 --no-entry --test-ca & # run some additional nodes
lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10003 --no-entry --test-ca &
lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10004 --no-entry --test-ca &
sleep 1
lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10005 --test-ca \
   "lunatic-rs/target/wasm32-wasi/release/examples/distributed.wasm" # run example from lunatic Rust binding repo
```

The next step is to implement multiple streams and chunking of messages to enable bandwidth control.


withtypes:
@bkolobara we could do something like that too. This approach is ok when you manually run nodes (you first create keys and then upload them after authenticating), but when provisioning servers and nodes, you need to automate this. Our plan was to generate a short-lived (ephemeral) token on the control server, and then you use it when the node starts. The token is used to authenticate only the first time, afterwards keys are generated and used for connections.

We were discussing how to simplify things for local testing.

theduke:
Drive by comment since I'm subscribed:

No need for openssl, I solved this with [rcgen](https://docs.rs/rcgen/latest/rcgen/)

```
fn generate_certificates(
    cert_path: &Path,
    key_path: &Path,
    domains: Vec<String>,
) -> Result<(), anyhow::Error> {
    let build = rcgen::generate_simple_self_signed(domains)?;
    let cert_raw = build.serialize_pem()?;
    let key_raw = build.serialize_private_key_pem();

    std::fs::write(cert_path, cert_raw)?;
    std::fs::write(key_path, key_raw)?;

    Ok(())
}
```

crates/lunatic-distributed/src/control/server.rs
```diff
@@ -54,8 +57,10 @@ impl Server {
 
     fn register(&self, reg: Registration) -> Response {
         let node_id = self.next_node_id();
+        let sign_request = CertificateSigningRequest::from_pem(&reg.signing_request).unwrap();
+        let signed_cert = sign_node_cert(&self.inner.ca_cert, sign_request).unwrap();
         self.inner.nodes.insert(node_id, reg);
-        Response::Register(node_id)
+        Response::Register((node_id, signed_cert))
```
withtypes, line 63: Use named fields, not tuples.

crates/lunatic-distributed/src/control/server.rs
```diff
@@ -54,8 +57,10 @@ impl Server {
 
     fn register(&self, reg: Registration) -> Response {
         let node_id = self.next_node_id();
+        let sign_request = CertificateSigningRequest::from_pem(&reg.signing_request).unwrap();
+        let signed_cert = sign_node_cert(&self.inner.ca_cert, sign_request).unwrap();
```
withtypes, line 61: unwraps?

crates/lunatic-distributed/src/control/server.rs
```diff
@@ -79,15 +84,59 @@ impl Server {
     }
 }
 
-impl Default for Server {
-    fn default() -> Self {
-        Self::new()
+pub static CTRL_SERVER_NAME: &'static str = "ctrl.lunatic.cloud";
+pub static ROOT_CERT: &'static str = include_str!("../../certs/root.pem");
```
withtypes, line 88: Let's name it `TEST_ROOT_CERT` and `TEST_ROOT_KEYS`.

crates/lunatic-distributed/src/control/server.rs
```diff
@@ -79,15 +84,59 @@ impl Server {
...
+    ca_cert: Option<&str>,
+    ca_keys: Option<&str>,
+) -> Result<Certificate> {
+    if test_ca {
+        let key_pair = KeyPair::from_pem(ROOT_KEYS)?;
+        let root_params = CertificateParams::from_ca_cert_pem(ROOT_CERT, key_pair)?;
+        let root_cert = Certificate::from_params(root_params)?;
+        Ok(root_cert)
+    } else {
+        let ca_cert_pem = std::fs::read(Path::new(ca_cert.unwrap()))?;
```
withtypes, line 102: Unwrap on option... use `ok_or_else` and `?` https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

crates/lunatic-distributed/src/control/server.rs
```diff
@@ -79,15 +84,59 @@ impl Server {
...
     }
 }
 
-pub async fn control_server(socket: SocketAddr, cert: String, key: String) -> Result<()> {
-    let mut quic_server = new_quic_server(socket, cert, key)?;
-    let server = Server::new();
+fn sign_node_cert(ca: &Certificate, sign_request: CertificateSigningRequest) -> Result<String> {
+    sign_request
+        .serialize_pem_with_signer(ca)
+        .map_err(|_| anyhow::anyhow!("Error while signing node certificate."))
```
withtypes, line 115: Does the error returned by `serialize_pem_with_signer` has some meaningful info?

kosticmarin: I've added an `Error(String)` response for [RcgenError](https://docs.rs/rcgen/0.9.3/rcgen/enum.RcgenError.html) which implements display.

withtypes (approved):

bkolobara:
I'm trying to test this a bit. When I run this command the first time:
```
lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10005 --test-ca \
   "lunatic-rs/target/wasm32-wasi/release/examples/distributed.wasm"
```
everything works fine, but on the second run I get:
```
[2022-08-01T14:29:01Z WARN  lunatic::mode::execution] Do not use test Certificate Authority in production!
[2022-08-01T14:29:01Z WARN  rustls::conn] Sending fatal alert BadCertificate
[2022-08-01T14:29:01Z WARN  lunatic_process] Process 1 failed, notifying: 0 links 
                            (Set ENV variable `RUST_LOG=lunatic=debug` to show stacktrace)
```
This could also be related to the fact that a node doesn't deregister itself when exiting.

bkolobara:
I'm having trouble running some examples. The following fails to run in a distributed environment:
```rust
use lunatic::{Mailbox, Process};

#[lunatic::main]
fn main(mailbox: Mailbox<i32>) {
    let nodes = lunatic::distributed::nodes();
    println!("Nodes {nodes:?}");

    let this = mailbox.this();

    nodes.into_iter().for_each(|node| {
        let remote = Process::spawn_node(node, this.clone(), hello);
        for i in 0..=1_000_000 {
            remote.send(i);
            let echo = mailbox.receive();
            assert_eq!(echo, i)
        }
    });
}

fn hello(parent: Process<i32>, mailbox: Mailbox<i32>) {
    loop {
        let v = mailbox.receive();
        parent.send(v);
        println!("{v}");
    }
}
```

The remote node fails:
```
# lunatic --control 127.0.0.1:10410 --control-server --no-entry --test-ca --node 127.0.0.1:10416

[2022-08-01T14:47:54Z WARN  lunatic::mode::execution] Do not use test Certificate Authority in production!
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: DeserializationFailed(Bincode(Io(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })))', /Users/bkolobara/dev/rust-lib/src/mailbox.rs:42:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2022-08-01T14:48:02Z WARN  lunatic_process] Process 1 failed, notifying: 0 links 
                            (Set ENV variable `RUST_LOG=lunatic=debug` to show stacktrace)
```

bkolobara:
I'm also running in trouble with the order of messages. This code:
```rust
use lunatic::{Mailbox, Process};

fn main() {
    let nodes = lunatic::distributed::nodes();
    println!("Nodes {nodes:?}");

    nodes.into_iter().for_each(|node| {
        let remote = Process::spawn_node(node, "First string".to_owned(), hello);
        remote.send("Second string".to_owned());
    });
}

fn hello(first: String, mailbox: Mailbox<String>) {
    println!("{first}");
    println!("{}", mailbox.receive());
}
```

Prints out on the local node:
```rust
First string
Second string
```

But on remote ones it displays:
```rust
Second string
First string
```

Order of messages between 2 processes should be guaranteed.

bkolobara:
I'm also getting a lot of random issues depending on what types I send in what order. E.g.
```
use std::time::Duration;

use lunatic::{sleep, Mailbox, Process};

#[lunatic::main]
fn main(mailbox: Mailbox<i32>) {
    let nodes = lunatic::distributed::nodes();
    println!("Nodes {nodes:?}");

    let this = mailbox.this();

    nodes.into_iter().for_each(|node| {
        let remote = Process::spawn_node(node, this.clone(), hello);
        remote.send("Some string data".to_owned());
    });
    sleep(Duration::from_millis(20000));
}

fn hello(_parent: Process<i32>, mailbox: Mailbox<String>) {
    loop {
        let v = mailbox.receive();
        println!("{v}");
    }
}
```
remote node receives an empty string?? But:
```
use std::time::Duration;

use lunatic::{sleep, Mailbox, Process};

#[lunatic::main]
fn main(_: Mailbox<i32>) {
    let nodes = lunatic::distributed::nodes();
    println!("Nodes {nodes:?}");

    nodes.into_iter().for_each(|node| {
        let remote = Process::spawn_node(node, (), hello);
        remote.send("Some string data".to_owned());
    });
    sleep(Duration::from_millis(20000));
}

fn hello(_parent: (), mailbox: Mailbox<String>) {
    loop {
        let v = mailbox.receive();
        println!("{v}");
    }
}
```
works as expected (remote node receives "Some string data").

bkolobara:
All issues happen with both implementations for me (s2n & quinn).

withtypes:
> I'm trying to test this a bit. When I run this command the first time:
> 
> ```
> lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10005 --test-ca \
>    "lunatic-rs/target/wasm32-wasi/release/examples/distributed.wasm"
> ```
> 
> everything works fine, but on the second run I get:
> 
> ```
> [2022-08-01T14:29:01Z WARN  lunatic::mode::execution] Do not use test Certificate Authority in production!
> [2022-08-01T14:29:01Z WARN  rustls::conn] Sending fatal alert BadCertificate
> [2022-08-01T14:29:01Z WARN  lunatic_process] Process 1 failed, notifying: 0 links 
>                             (Set ENV variable `RUST_LOG=lunatic=debug` to show stacktrace)
> ```
> 
> This could also be related to the fact that a node doesn't deregister itself when exiting. It just makes testing a pain, because I need to keep changing the ip all the time.

@bkolobara this happens because we generate a unique name for each node for which the cert is valid. When you restart, it generates a different name, but it listens on the same socket, so connecting to the old one gives you "bad cert". Deregistering would help but if you reconnect quickly or your node didn't refresh the list of all nodes, it would happen again. It's a bit more complicated, so we kept it simple for now.

I will add for now (should be easy):

- remove the old node with the same socket address
- when the connection to control drops, remove it (but we need to think about reconnects as we want to keep the node id/name... it might be a network problem)

By the way, both Marin and I test with a script we provided and `killall lunatic`.

withtypes:
Regarding the order of messages 🤦‍♂️ as we implemented the control server first, we repeated the same logic: spawn a task on each message, but we actually need to process them sequentially on the node server (coming from another node). It's an easy fix: replacing `spawn` with `.await` in the connection read loop. I will fix it and test your examples.

withtypes:
@bkolobara I've also added a simple deregisteration (little bit clunky as we don't track connection node ids, so node will send its own id for now). I've also temporarily added removal of nodes which have conflicting address until we define reconnects, dead connections, etc. This PR is just MVP with decoupled distributed, new control node architecture, and now QUIC connections.

See 09e1ca7f5131dbcf5df0f51429f3f51493564c13

withtypes:
By the way, here is the full script which I use in the root directory containing `lunatic` and `lunatic-rs` repos. It rebuilds both vm and lib, and restarts all nodes.

```bash
#!/bin/bash
echo -ne '\033c' # clear scroll-back in terminal
set -e
SCRIPT=$(readlink -f "$0")
BASEDIR=$(dirname "$SCRIPT")

export RUST_LOG="warn,lunatic=info,lunatic_process=trace,lunatic_distributed_api=debug"

cd $BASEDIR/lunatic
cargo build --release --quiet

cd $BASEDIR/lunatic-rs
cargo build --release --examples --quiet

prefix() {
    # A helper to add prefix to each stdout/stderr line. It's useful to tell nodes apart
    local prefix="$1"
    shift
    echo "RUN $prefix: $@"
    "$@" > >(sed "s/^/$prefix: /") 2> >(sed "s/^/$prefix: /" >&2)
}

lunatic=$BASEDIR/lunatic/target/release/lunatic
CERTS=$BASEDIR/keygen/certs

killall lunatic || true

prefix "CTRL" $lunatic --control 127.0.0.1:10410 --control-server --no-entry --test-ca &
sleep 1
prefix "A" $lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10002 --no-entry --test-ca &
prefix "B" $lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10003 --no-entry --test-ca &
#prefix "C" $lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10004 --no-entry --test-ca &
sleep 1
prefix "MAIN" $lunatic --control 127.0.0.1:10410  --node 127.0.0.1:10005 --test-ca "$BASEDIR/lunatic-rs/target/wasm32-wasi/release/examples/distributed2.wasm"
```

bkolobara:
Thanks, this makes testing much easier. Maybe we should set up a `integration-testing` repository with all kinds of custom setups and tests. We could also pull it in as part of the CI.

withtypes:
Good idea, we need to do something like this as testing multi-node, control, bandwidth/latency benchmarks, is beyond unit-tests and Rust tools.

PR #118, opened 2022-07-26
Migrate lunatic-timer-api from async-std to tokio

MarkintoshZ:
Migrating the last part of the code base that still uses async-std to tokio.

- [x] Build passed all tests in lunatic-rs.


bkolobara:
Looks like we forgot to port this API over to the new tokio stuff. Good catch @MarkintoshZ!

PR #129, opened 2022-09-04
Fix confusing error output

HurricanKai:
I'm still not even sure how I ended up here, but certainly quite confusing


bkolobara:
Thanks for the fix!

PR #131, opened 2022-09-07
Feature/distributed metadata

### LINKED ISSUE #121: Naming distributed nodes
We should be able to name distributed nodes to easier differentiate them. This would also include host function for looking up nodes by name and returning names when getting all nodes.

kosticmarin:
Resolves #121 

This PR adds a new argument when starting nodes called `tag` which can repeat. The `tag` argument is used
to store additional user defined metadata about the node in a form of a `key=value`, e.g. `--tag name=node01`.

Further we add a two new functions for looking up nodes via a query. The query syntax is like [URL query string](https://en.wikipedia.org/wiki/Query_string).
The first function `exec_lookup_nodes` executes the query on the control node and stores the results in the 
control client, returning a query id and query results size.
The second function `copy_lookup_nodes_results` copies the query results to the guest memory based on the query id.

Finally on the control server we add a parser for the query string.

### DIFF
--- a/crates/lunatic-distributed/src/control/message.rs
+++ b/crates/lunatic-distributed/src/control/message.rs
@@ -1,6 +1,6 @@
-use std::{collections::HashMap, net::SocketAddr};
 use bytes::Bytes;
 use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, net::SocketAddr};
 
 use crate::NodeInfo;
 


crates/lunatic-distributed/src/control/client.rs
```diff
@@ -23,19 +24,23 @@ pub struct Client {
 
 pub struct InnerClient {
     next_message_id: AtomicU64,
+    next_query_id: AtomicU64,
     node_addr: SocketAddr,
     node_name: String,
     control_addr: SocketAddr,
     connection: Connection,
     pending_requests: DashMap<u64, Arc<AsyncCell<Response>>>,
+    pending_queries: DashMap<u64, Vec<u64>>,
```
withtypes, line 33: Why "pending", these are resolved queries. Let's simply call it `node_queries`?

crates/lunatic-distributed/src/control/client.rs
```diff
@@ -23,19 +24,23 @@ pub struct Client {
...
     nodes: DashMap<u64, NodeInfo>,
     node_ids: RwLock<Vec<u64>>,
+    node_metadata: HashMap<String, String>,
 }
 
 impl Client {
     pub async fn register(
         node_addr: SocketAddr,
         node_name: String,
+        node_metadata: HashMap<String, String>,
```
withtypes, line 43: Maybe simply call it node "attributes".

crates/lunatic-distributed/src/control/parser.rs
```diff
@@ -0,0 +1,323 @@
+use anyhow::{anyhow, Result};
+use dashmap::mapref::multiple::RefMulti;
+
```
withtypes, line 3: Do we have some description of this language? What are valid attribute key/values? What operators?

src/mode/execution.rs
```diff
@@ -18,6 +18,22 @@ use lunatic_runtime::{DefaultProcessConfig, DefaultProcessState};
 
 use uuid::Uuid;
 
+/// Parse a single key-value pair
+fn parse_key_val<K, V>(s: &str) -> Result<(K, V)>
```
withtypes, line 22: This doesn't use the parser?

withtypes (commented):
Looks ok to me. It adds two new guest fns, it's backwards compatible. @bkolobara we can merge it for next release if it looks good to you.

withtypes (approved):

bkolobara:
Could you just fix the issues after merging the other branch into main and I'm going to merge this one too?

kosticmarin:
@bkolobara Merge issues have been resolved.

bkolobara:
Awesome! Thanks!

PR #132, opened 2022-09-11
Make it more explicit that value must be 16-byte.

teymour-aldridge:


bkolobara:
Thanks for clarifying and improving the docs!

PR #137, opened 2022-09-25
Add Metrics to VM

HurricanKai:
This adds a variety of metrics to a bunch of methods, using the [metrics-rs](https://github.com/metrics-rs/metrics) crate.
According to the crate, as long as no recorder is installed, this has little performance impact.

All code is behind the "metrics" feature flag, but I have enabled it by default.
No recorder is included by default, but the "prometheus" feature flag adds the prometheus exporter, with appropriate CLI arguments.

I've tested this with [frenezulo](https://github.com/HurricanKai/frenezulo) and found the results to be quite useful already.
I plan to expose tokio metrics in the future & potentially allow WASM clients to expose their own metrics too, not sure yet.

[example prometheus output](https://github.com/lunatic-solutions/lunatic/files/9640804/sample.txt)

In my opinion the prometheus exporter could be included by default, but it should not be enabled (using `--prometheus` currently) if the user doesn't use it, as histograms appear to be tracked indefinitely, so if no one ever scrapes the prometheus output, it's memory usage will grow until the process is killed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1418,6 +1418,7 @@ dependencies = [
  "lunatic-process",
  "lunatic-stdout-capture",
  "lunatic-wasi-api",
+ "metrics",
  "tokio",
  "wasmtime",
 ]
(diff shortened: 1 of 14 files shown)


bkolobara:
Looks great. Thanks for the contribution!

PR #147, opened 2022-10-18
Feat: Add(sqlite)

jtenner:
This pull request aims to add sqlite capabilities to the lunatic host. Any comments and help would be appreciated!

### DIFF
--- a/crates/lunatic-sqlite-api/src/lib.rs
+++ b/crates/lunatic-sqlite-api/src/lib.rs
@@ -1,11 +1,11 @@
-use hash_map_id::HashMapId;
-use sqlite::Connection;
 use anyhow::Result;
-use wasmtime::{Linker, Caller, Trap};
-use lunatic_process::{state::ProcessState};
+use hash_map_id::HashMapId;
 use lunatic_common_api::{get_memory, IntoTrap};
 use lunatic_error_api::ErrorCtx;
+use lunatic_process::state::ProcessState;
+use sqlite::Connection;
 use std::sync::Mutex;
+use wasmtime::{Caller, Linker, Trap};
 
 pub type SQLiteResource = HashMapId<Mutex<Connection>>;
 
@@ -26,7 +26,7 @@ fn open<T: ProcessState + ErrorCtx + SQLiteCtx>(
     mut caller: Caller<T>,
     path_str_ptr: u32,
     path_str_len: u32,
-    connection_id_ptr: u32
+    connection_id_ptr: u32,
 ) -> Result<u32, Trap> {
     let memory = get_memory(&mut caller)?;
     let (memory_slice, _state) = memory.data_and_store_mut(&mut caller);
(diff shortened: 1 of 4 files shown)


crates/lunatic-sqlite-api/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+    let exec = std::str::from_utf8(exec).or_trap("lunatic::sqlite::execute")?;
+
+    // execute a single sqlite query
+    state
+        .sqlite_resources()
+        .get(conn_id)
+        .or_trap("lunatic::sqlite::execute")?
+        .lock()
+        .or_trap("lunatic::sqlite::execute")?
+        .execute(exec)
```
bkolobara, line 87: It would be nice to return here the error to the guest, instead of just trapping.

crates/lunatic-sqlite-api/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+        .or_trap("lunatic::sqlite::execute")
+}
+
+fn query<T: ProcessState + ErrorCtx + SQLiteCtx>(
+    mut caller: Caller<T>,
+    conn_id: u64,
+    query_str_ptr: u32,
+    query_str_len: u32,
+    data_ptr: u32,
+    data_len: u32,
```
bkolobara, line 97: Instead of `query` taking the `data_*`, I think it would be better if it returned just the length and buffered the data. There could be another function (`query_result`?) that just copies the data into the guest.

That way the guest has a chance to allocate the right space and doesn't need to hope that the data is going to fit into a pre-allocated buffer.

jtenner: I was originally for this idea. We can talk about it on Monday for sure.

crates/lunatic-sqlite-api/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+        let count = statement.column_count();
+        for i in 0..count {
+            let mut bytes = match statement.column_type(i) {
+                sqlite::Type::Binary => {
+                    let mut bytes = statement
+                        .read::<Vec<u8>>(i)
+                        .or_trap("lunatic::sqlite::query_prepare::read_binary")?;
+
+                    let len = bytes.len();
+                    let mut result = vec![ColumnType::Binary as u8];
```
bkolobara, line 140: I think this also needs to be `.to_le_bytes` encoded.

crates/lunatic-sqlite-api/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+                sqlite::Type::String => {
+                    let bytes = statement
+                        .read::<String>(i)
+                        .or_trap("lunatic::sqlite::query_prepare::read_string")?;
+
+                    let len = bytes.len();
+                    let mut result = vec![ColumnType::String as u8];
+                    result.append(&mut (len as u32).to_le_bytes().to_vec());
+                    result.append(&mut bytes.as_bytes().to_vec());
+                    result
```
bkolobara, line 176: Instead of appending to a temporary vector (`result`), we could here directly append to `return_value`.

crates/lunatic-sqlite-api/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+                }
+                sqlite::Type::Null => vec![ColumnType::Null as u8],
+            };
+            return_value.append(&mut bytes);
+        }
+
+        return_value.push(ColumnType::NewRow as u8);
```
bkolobara, line 183: This is going to leave a trailing `NewRow` marker, even there are no new rows. This is not a big issue, but the parser needs to be aware of it.

crates/lunatic-sqlite-api/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+    mut caller: Caller<T>,
+    resource_id: u64,
+    data_ptr: u32,
+    data_len: u32,
+) -> Result<(), Trap> {
+    // get the memory and the state
+    let memory = get_memory(&mut caller)?;
+    let (memory_slice, state) = memory.data_and_store_mut(&mut caller);
+
+    // get the vevtor
```
bkolobara, line 220: typo

bkolobara:
This approach looks good to me. Thanks @jtenner for pushing this effort forward!

I'm not that familiar with SQLite and am curious how useful features like "named parameters" are? At the moment we don't support something like:
```rust
let mut statement = connection.prepare("SELECT * FROM users WHERE name = :name")?;
statement.bind_by_name(":name", "Bob")?;
```
Should this also be part of the interface, or are people just working around this? I assume that doing manual escaping of user input without it would be an error-prone process. Should we include it in our API?

I'm also wondering about the security implications of the `open` call. Currently the vm only lets you access the current directory and sub-directories, but the `open` command here takes **any** path. Should this be configurable like directory access? And limited by default?

On a more future-proof note, tools like https://litestream.io that work together with SQLite often use the WAL (write ahead log) feature. This is also recommended for all production deployments of SQLite. If I'm not mistaken, this also allows for multithreaded writes. Should we maybe use it by default? This would allow for connections to be shared between processes running concurrently.

withtypes:
By the way, could this work as a plugin as in #154 ? Also @tqwewe

tqwewe:
Absolutely. I haven't followed along with this PR, but after looking into it, it seems like it just adds host functions and stores state?
If so, then it'sa perfect fit for a plugin.

jtenner:
Closed in favor of #160

PR #148, opened 2022-10-21
Shared compiled wasm modules v2

tqwewe:
Allows compiled wasm modules to be (de)serialized and shared.

Replaces PR https://github.com/lunatic-solutions/lunatic/pull/126. This PR does not remove pre-linking wasm modules, and the performance has not changed compared to main branch.

```
Main branch: Avg 47.98 µs

spawn process           time:   [44.157 µs 44.462 µs 44.791 µs]
spawn process           time:   [52.646 µs 53.629 µs 54.587 µs]
spawn process           time:   [49.791 µs 51.060 µs 52.506 µs]
spawn process           time:   [43.909 µs 44.194 µs 44.514 µs]
spawn process           time:   [46.084 µs 46.548 µs 47.066 µs]


Shared Module: Avg 47.14 µs

spawn process           time:   [45.188 µs 45.432 µs 45.725 µs]
spawn process           time:   [44.663 µs 44.896 µs 45.159 µs]
spawn process           time:   [53.761 µs 54.157 µs 54.575 µs]
spawn process           time:   [45.544 µs 45.771 µs 46.025 µs]
spawn process           time:   [44.897 µs 45.425 µs 46.307 µs]
```

I believe this might be what you meant by your comment @bkolobara 
> We only have one state (generic T) though. Could we bubble it up all the way to Resource<T> and then use the state type to create the specific Resource?
<https://github.com/lunatic-solutions/lunatic/pull/126#issuecomment-1220862816>

Once this is approved and merged, https://github.com/lunatic-solutions/lunatic-rs/pull/68 can also be merged.

### DIFF
--- a/crates/lunatic-process/src/message.rs
+++ b/crates/lunatic-process/src/message.rs
@@ -10,7 +10,7 @@ use std::{
     sync::Arc,
 };
 
-use lunatic_networking_api::TcpConnection;
+use lunatic_networking_api::{TcpConnection, TlsConnection};
 use tokio::net::UdpSocket;
 
 use crate::{runtimes::wasmtime::WasmtimeCompiledModule, Process};
@@ -33,6 +33,16 @@ impl<T> Message<T> {
         match self {
             Message::Data(message) => message.tag,
             Message::LinkDied(tag) => *tag,
+        }
+    }
+
+    #[cfg(feature = "metrics")]
+    pub fn write_metrics(&self) {
+        match self {
+            Message::Data(message) => message.write_metrics(),
+            Message::LinkDied(_) => {
+                metrics::increment_counter!("lunatic.process.messages.link_died.count");
+            }
         }
     }
 }
@@ -91,6 +101,12 @@ impl<T> DataMessage<T> {
     /// Adds a UDP socket to the message and returns the index of it inside of the message
     pub fn add_udp_socket(&mut self, udp_socket: Arc<UdpSocket>) -> usize {
         self.resources.push(Resource::UdpSocket(udp_socket));
+        self.resources.len() - 1
+    }
+
+    /// Adds a TLS stream to the message and returns the index of it inside of the message
+    pub fn add_tls_stream(&mut self, tls_stream: Arc<TlsConnection>) -> usize {
+        self.resources.push(Resource::TlsStream(tls_stream));
         self.resources.len() - 1
     }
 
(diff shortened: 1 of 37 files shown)


crates/lunatic-process/src/message.rs
```diff
@@ -233,19 +259,21 @@ impl Read for DataMessage {
...
 /// ...) that is attached to a [`DataMessage`].
-pub enum Resource {
+pub enum Resource<T> {
     None,
-    Process(Arc<dyn Process>),
+    Process(Arc<dyn Process<T>>),
+    Module(Arc<WasmtimeCompiledModule<T>>),
     TcpStream(Arc<TcpConnection>),
     UdpSocket(Arc<UdpSocket>),
     TlsStream(Arc<TlsConnection>),
```
bkolobara, line 268: The `T` generic spreads really everywhere. I assume it's mostly because of Messages/Signals getting a `T`, because they can contain a `Resource<T>`.

Would it be possible just to do `Arc<dyn Any>` for resources and then have the `take_*` functions have a generic, so they try to downcast to the right type and if not just return `None`. The resources vector would be more of a `Vec<Option<Arc<dyn Any>>>`. All of the resources are already wrapped into a `Arc`. This could even simplify future additions to resources, containing more generics.

tqwewe: That cleaned things up quite a bit. I've never used dyn Any before really, but came in handy here.

The only downside is I wasn't able to get it working with `fn take_process() -> Arc<dyn Process>` since I can't downcast into `dyn Process`. The `take_process` function though is actually not used anywhere throughout the project, so it doesn't seem to be much of an issue for now.

bkolobara: Yeah, that can be removed. Processes are now addressable through the ID and don't need to be put as resources into messages.

This change reduced the added lines by 50%! Looks great! I will merge it as soon as you remove the commented out `take_process` stuff.

bkolobara: It looks like it was removed already, my bad. I looked at the wrong diff.

tqwewe: I'm not quite sure which part you're referring to? You might have viewed before this commit: https://github.com/lunatic-solutions/lunatic/pull/148/commits/7c9201351f038d8ac330d200ef9eaa3e9a1133b7 ?

PR #152, opened 2022-11-10
Implement a Stacked Scratch Area

jtenner:
This allows for messages to be stacked. Calling create_message() adds a message onto the stack, and any message modificatiosn modify the message on the top of the stack.

This is an experiment per a discussion I had with @bkolobara .

### DIFF
--- a/crates/lunatic-distributed-api/src/lib.rs
+++ b/crates/lunatic-distributed-api/src/lib.rs
@@ -210,10 +210,13 @@ where
     for<'a> &'a T: Send,
 {
     Box::new(async move {
-        let message = caller
+        let mut stack = caller
             .data_mut()
             .message_scratch_area()
             .take()
+            .or_trap("lunatic::message::send::no_message")?;
+        let message = stack 
+            .pop()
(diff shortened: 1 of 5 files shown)


tqwewe:
This is interesting, whats the main benefit of this, is it to be able to store multiple thingsi n the scratch area?

withtypes:
I also played with similar ideas:

- you create a new message and receive a handle that you can use (write into it, drop it, send it once or multiple times)
- you receive a message and receive a handle and you can decide to read only a few bytes, to forward it, or drop it when you don't want to keep it in lunatic memory anymore

bkolobara:
There could also be an opportunity for a bigger abstraction here. Instead of message handles, we could be creating general buffers in the vm. They could be sent as part of messages, but also be used for other stuff, like "read a tcp stream into a buffer, one process checks just a few bytes (path) and dispatches it to a handler".

jtenner:
I don't know what the right decision is, but I want to revisit this further. It would involve refactoring the entire message api altogether.

PR #154, opened 2022-11-15
Native plugins support

tqwewe:
Adds support for native plugins (.dll/.so/.dylib) which can be loaded with the `--plugins` flag.

An example exists in the `examples/` directory, along with some docs on running the example.

Related to https://github.com/lunatic-solutions/lunatic/issues/69 *(nice)*


src/state.rs
```diff
@@ -131,6 +138,8 @@ impl ProcessState for DefaultProcessState {
             wasi_stderr: None,
             initialized: false,
             registry: self.registry.clone(),
+            plugins: self.plugins.clone(),
+            plugin_state: HashMap::new(),
```
tqwewe, line 142: Should we clone the plugin state here? Or put the whole plugin state in an Arc?
The only problem is that we cannot `downcast_mut` when the state is in an Arc, or even borrow it mutably.

src/state.rs
```diff
@@ -456,11 +477,31 @@ impl DistributedCtx<LunaticEnvironment> for DefaultProcessState {
             wasi_stderr: None,
             initialized: false,
             registry: Default::default(), // TODO move registry into env?
+            plugins: Arc::new(vec![]),
+            plugin_state: HashMap::new(),
```
tqwewe, line 481: Should plugins and plugin state be empty in `new_dist_state`?

kosticmarin: Probably not (I'm not sure :smile: ), when reading through #69 the original idea was to have different plugins for each Environment.

> Testing plugins or comparing performance between competing implementations becomes trivial, just spawn different environments, attach different plugins to them and run the same code. 

When spawning via distributed API we use Environment to crate state.

If I understand this correctly calling a function from a dynamic library will fail, in this case when spawning a process on a different node, since plugin is not linked?

tqwewe:
I just came across an issue, and I'm not sure whats going on.

Screen Shot 2022-11-17 at 12 05 15 pm

When I run lunatic in release, everything works fine... but running lunatic in debug mode causes the host functions to silently fail 🤔

bkolobara:
Sorry, it took me a bit longer to get around and check this out. Great work so far!

I also can't figure out what the issue with release/debug builds could be. From my understanding on how dynamic loading works, it shouldn't be an issue. It might be something Wasmtime specific, because they do some assembly magic to allow stack switching? Maybe creating a minimal reproducible example with just wasmtime and dynamic loading could help us to pinpoint it.

scutuatua-crypto (approved):

PR #156, opened 2022-11-16
Format Cargo.toml files

tqwewe:
Sorts dependencies alphabetically, and places local crates at the top, and external crates below. Blesses project with some OCD.


withtypes:
Maybe we can use this opportunity to use cargo 1.64 feature of workspace versions?

- [Cargo release notes](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-164-2022-09-22)
- [Cargo issue](https://github.com/rust-lang/cargo/issues/3931)
- [how wasmtime uses it](https://github.com/bytecodealliance/wasmtime/blob/main/crates/runtime/Cargo.toml) e.g. `wasmtime-environ = { workspace = true }`

tqwewe:
Great idea! I've pushed some commits with this change.
I also manually went through each project to find which dependencies and features are unused, and based on the Cargo.lock, I was able to get rid of `wasmparser` and `futures-macro` which were both unused.

PR #163, opened 2022-11-30
add process::exists

jtenner:
This adds a host function to help the guest determine if a process is still alive.


bkolobara:
Thanks for contributing this!

PR #164, opened 2022-12-01
add metrics API

Roger:
Expose metrics-rs APIs to wasm modules

Would be nice to have also https://github.com/lunatic-solutions/lunatic/pull/159 merged, so metrics can be shown in prometheus


bkolobara:
Looks awesome. Thanks for the contribution @Roger! <3

PR #166, opened 2022-12-05
Implement local control server

teskeras:


crates/lunatic-control-axum/src/routes.rs
```diff
@@ -0,0 +1,155 @@
...
+    log::info!("Registration for node name {}", reg.node_name);
+
+    let control = control.as_ref();
+    let cert_pem = CertificateSigningRequest::from_pem(&reg.csr_pem)
+        .and_then(|sign_request| sign_request.serialize_pem_with_signer(&control.ca_cert))
+        .map_err(|e| ApiError::custom("sign_error", e.to_string()))?;
+
+    let mut authentication_token = [0u8; 32];
+    getrandom::getrandom(&mut authentication_token)
+        .map_err(|e| ApiError::log_internal("Error generating random token fin registration", e))?;
```
withtypes, line 32: for - typo

crates/lunatic-distributed/Cargo.toml
```diff
@@ -18,7 +18,7 @@ bytes = "1"
 dashmap = { workspace = true }
 log = { workspace = true }
 quinn = { version = "0.9" }
-rcgen = { version = "0.10", features = ["pem", "x509-parser"] }
+rcgen = { workspace = true }
```
withtypes, line 21: We had a discussion about which crates should be workspace defined. Maybe rcgen shoudn't as it's used only in distributed and for PEM format shouln't matter which version/lib generated it.

PR #167, opened 2022-12-06
upgrade wasmtime to 3.0

jtenner:
Tests pass locally, everything seems to work. Ready for review!


crates/lunatic-error-api/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use hash_map_id::HashMapId;
 use lunatic_common_api::{get_memory, IntoTrap};
-use wasmtime::Trap;
+
```
bkolobara, line 4: Don't need the empty space.

crates/lunatic-process/src/runtimes/wasmtime.rs
```diff
@@ -155,15 +155,7 @@ where
...
-                        Some(trap) => {
-                            if trap.i32_exit_status().is_some()
-                                && trap.i32_exit_status().unwrap() == 0
-                            {
-                                ResultValue::Ok
-                            } else {
-                                ResultValue::Failed(trap.to_string())
-                            }
-                        }
+                        Some(trap) => ResultValue::Failed(trap.to_string()),
```
bkolobara, line 158: You will still need to check for an exit code `0`, but instead of using `wasmtime:Trap` you will need to use: https://docs.rs/wasmtime-wasi/3.0.1/wasmtime_wasi/struct.I32Exit.html#impl-Pointee-for-I32Exit

crates/lunatic-registry-api/src/lib.rs
```diff
@@ -41,7 +40,7 @@ pub fn register<T: ProcessState + ProcessCtx<T> + 'static>(linker: &mut Linker<T
 
 // Registers process with ID under `name`.
 //
-// Traps:
+// s:
```
bkolobara, line 43: Typo

crates/lunatic-stdout-capture/Cargo.toml
```diff
@@ -8,5 +8,5 @@ repository = "https://github.com/lunatic-solutions/lunatic/tree/main/crates/luna
 license = "Apache-2.0/MIT"
 
 [dependencies]
-wasi-common = "2"
-wiggle = "2"
+wasi-common = "3"
+wiggle = "3"
```
bkolobara, line 12: I think we should use here workspace decencies like in all the other crates.

bkolobara (changes requested):
Looks good to me, just requires a few small changes and can be merged.

crates/lunatic-process/src/runtimes/wasmtime.rs
jtenner: I updated this based on the conversation in discord.

PR #170, opened 2022-12-21
Feature/new ctrl

withtypes:
The biggest change in this PR is using HTTP for the control server, moving from our custom request/response protocol based on QUIC. Because nodes use QUIC to communicate, originally the idea was to use the same protocol for node-control messages. The main advantage is that HTTP is well supported and understood by devs so it's easier to implement custom control servers using different languages (lunatic sms?). @kosticmarin and @teskeras contributed greatly to this PR. 👍 

# New Control

Besides moving to HTTP, we have separated start into two parts:

1. on the first run, node will register itself using a provided URL (possibly with a short-lived authentication token). It will generate a cert signing request and later use mTLS for authentication (between nodes and for other control API endpoints)
2. whenever a node starts, it will notify the control server and receive a new node id

Other API endpoints mirror the previous messages: node list, get and add module. The current control server is implemented using Axum and stores everything in-memory.

# New CLI Arguments

I have changed the CLI arguments to be based on commands:

- `lunatic run some.wasm` will run a wasm module as before without distributed lunatic
- `lunatic control` will start only a control server (see `lunatic help control` for other arguments). By default it starts it on localhost:3000 with test certificates to simplify testing
- `lunatic node` will start a node. To simplify testing we can provide a wasm module that will be run on this node with `lunatic node --wasm some.wasm`

# Other Changes

- mutual TLS
- ctrl-c handler for graceful stopping of nodes
- MessagePack instead of Bincode for communication (this might be temp as it was the simplest format which stays


crates/lunatic-control-axum/src/routes.rs
```diff
@@ -0,0 +1,154 @@
...
+    let control = control.as_ref();
+    let module_id = control.add_module(body.to_vec());
+    ok(ModuleId { module_id })
+}
+
+pub async fn get_module(
+    node_auth: NodeAuth,
+    PathExtractor(id): PathExtractor<u64>,
+    control: Extension<Arc<ControlServer>>,
+) -> ApiResponse<ModuleBytes> {
```
kosticmarin, line 131: This returns the module as json with `Vec<u8>`, we should probably return something like `body::Bytes`

withtypes: I agree but let's leave it for now, we can change it later.

PR #178, opened 2023-01-15
Run wasm default option

pinkforest:
- If subcommand is not found it is assumed it is `lunatic run <wasm>` without having to say `run`
- Allows backwards compatibility to 0.12
- Ergonomics: Less pain migrating to 0.13

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,7 +17,7 @@ fn is_run_implied() -> bool {
     // lunatic <foo.wasm> -> Implied run
     // lunatic run <foo.wasm> -> Explicit run
     // lunatic fdskl <foo.wasm> -> Not implied run
-    let test_re = Regex::new(r"^(--bench|--dir|[^\s]+\.wasm)")
+    let test_re = Regex::new(r"^(--bench|--dir|.+\.wasm)")
         .expect("BUG: Regex error with lunatic::mode::execution::is_run_implied()");
 
     test_re.is_match(&std::env::args().nth(1).unwrap())
--- a/src/main.rs
+++ b/src/main.rs
@@ -25,6 +25,15 @@ fn is_run_implied() -> bool {
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    // Run is implied from lunatic 0.12
+    let augmented_args = if is_run_implied() {
+        let mut augmented_args: VecDeque<String> = std::env::args().collect();
+        augmented_args.insert(1, "run".to_owned());
+        Some(augmented_args.into())
+    } else {
+        None
+    };
+
     // Detect if `cargo test` is running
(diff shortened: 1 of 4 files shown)


bkolobara:
Great work on figuring this out! There is also this part when running tests that now needs to receive an optional `run`: https://github.com/lunatic-solutions/lunatic/blob/main/src/mode/cargo_test.rs#L15-L18

Can this also be made optional?

pinkforest:
I found a better way - let me know if you liked the old better - now it handles --flags proper w/ the power of regex :)

**Now it's**

$ lunatic --bench
```
Usage: lunatic run --bench <PATH> [WASM_ARGS]...
```
$ lunatic --bench target/wasm32-wasi/debug/yaws.wasm
```
Listening on addr: 127.0.0.1:9999
```
$ lunatic --dir . target/wasm32-wasi/debug/yaws.wasm
```
Listening on addr: 127.0.0.1:9999
```
$ lunatic target/wasm32-wasi/debug/yaws.wasm
```
Listening on addr: 127.0.0.1:9999
```
$ lunatic errororor
```
error: The subcommand 'errororor' wasn't recognized

Usage: lunatic-runtime <COMMAND>

For more information, try '--help'.
```
$ lunatic --fff errororor
```
error: unexpected argument '--fff' found

Usage: lunatic <COMMAND>

For more information, try '--help'.
```

pinkforest:
> Can this also be made optional?

I adjusted the detection that was out there and now it works in both under cargo test and lunatic run :tada:

src/main.rs
```diff
@@ -3,10 +3,37 @@ mod mode;
...
+fn is_run_implied() -> bool {
+    if std::env::args().count() < 2 {
+        return false;
+    }
+
+    // lunatic <foo.wasm> -> Implied run
+    // lunatic run <foo.wasm> -> Explicit run
+    // lunatic fdskl <foo.wasm> -> Not implied run
+    let test_re = Regex::new(r"^(--bench|--dir|[^\s]+\.wasm)")
+        .expect("BUG: Regex error with lunatic::mode::execution::is_run_implied()");
```
bkolobara, line 21: Didn't get to try this out yet, but would this work with .wasm files that have spaces in the name?

pinkforest: Oh thanks - it does now :)

$ lunatic "foo bar.wasm"
```
Error: No such file or directory (os error 2)
```

bkolobara:
This looks good to me. My only worry is that we will need to remember to extend the regex every time we add an additional argument to `run`, but this shouldn't be a big issue and I'm going to merge this as is.

pinkforest:
>  remember to extend the regex every time we add an additional argument to run

Could always have a regex with ^- as no subcommand start with a dash

PR #179, opened 2023-01-16
Add new command

sid-707:
This PR adds a new command to Lunatic: `lunatic new <project-name>`.

It generates a new Lunatic Rust project with the name `<project-name>` by performing the following steps:
1) Run `cargo new <project_name>` to create the project.
2) Change the directory to `<project_name>`.
3) Add the latest version of Lunatic as a dependency to the project.
4) Run the `init` command to add the `.cargo/config.toml` file.
5) Replace Cargo's default `src/main.rs` with an example that uses Lunatic. 

The example program is a modified version of Bernard's example from [Writing Rust the Erlang way](https://youtu.be/NZ1rpQSLvg8?t=695).

### DIFF
--- a/src/mode/new.rs
+++ b/src/mode/new.rs
@@ -13,24 +13,30 @@ pub struct Args {
 }
 
 fn add_lunatic_main_file() {
-    let mut file = File::create("src/main.rs").expect("Could not open src/main.rs");
-    file.write_all(b"use std::time::Duration;\n\n\
-    use lunatic::{sleep, spawn_link};\n\n\
-    fn main() {\n\
-        spawn_link!(|| println!(\"Hello, world! I'm a process.\"));\n\
-        sleep(Duration::from_millis(100));\n\
-    }\n").expect("Could not write to src/main.rs");
+    let text = format!(
+        "use std::time::Duration;
+use lunatic::{{sleep, spawn_link}};
+        
+fn main() {{
+    spawn_link!(|| println!(\"Hello, World! I'm a process.\"));
+    sleep(Duration::from_millis(100));
+}}"
+    );
+
+    let mut file = File::create("src/main.rs").expect("Opened src/main.rs");
+
+    file.write_all(text.as_bytes()).expect("\"Hello, World!\" example written in src/main.rs");
 }
 
 pub(crate) fn start(args: Args) -> Result<()> {
     let project_name = &args.new_args[0];
 
-    Command::new("cargo").args(["new", project_name]).status().expect(format!("Failed to create {} project", project_name.as_str()).as_str());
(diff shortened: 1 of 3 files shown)


src/mode/new.rs
```diff
@@ -0,0 +1,52 @@
...
+        "use std::time::Duration;
+use lunatic::{{sleep, spawn_link}};
+        
+fn main() {{
+    spawn_link!(|| println!(\"Hello, World! I'm a process.\"));
+    sleep(Duration::from_millis(100));
+}}"
+    );
+
+    let mut file = File::create("src/main.rs").expect("Opened src/main.rs");
```
pinkforest, line 26: This is not portable. See https://doc.rust-lang.org/std/path/struct.PathBuf.html

src/mode/new.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[derive(Parser, Debug)]
+#[command(version)]
+pub struct Args {
+    #[arg(index = 1)]
+    pub new_args: Vec<String>,
+}
+
+fn add_lunatic_main_file() {
+    let text = format!(
```
pinkforest, line 16: Having it injected here makes maitenance difficult. a templating engine e.g. like https://crates.io/crates/askama could help.

mpope9:
Is there any status on merging this? I was considering adding a kind of scaffold like Phoenix has based off of my current HTTP server setup. It'd register a supervisor, a simple submillisecond HTTP server, register the server under the supervisor, _maybe_ a database connection and setup the `.cargo/config.toml` file.

bkolobara:
I'm a bit in doubt if we should keep extending the `lunatic` CLI with additional commands for setting up projects.

What I like at the moment is that you can easily just stick to `cargo` commands to start a new project and don't need to learn a new CLI tool. Should we maybe just extend cargo with `cargo submillisecond.new`, `cargo lunatic.new`, ...?

This would let us keep the runtime focused on more runtime specific "stuff", instead of also being a collection of starter project layouts.

sid-707:
> I'm a bit in doubt if we should keep extending the `lunatic` CLI with additional commands for setting up projects.
> 
> What I like at the moment is that you can easily just stick to `cargo` commands to start a new project and don't need to learn a new CLI tool. Should we maybe just extend cargo with `cargo submillisecond.new`, `cargo lunatic.new`, ...?
> 
> This would let us keep the runtime focused on more runtime specific "stuff", instead of also being a collection of starter project layouts.

I like the `cargo` approach more now that you mention it. It makes sense to let `lunatic` remain as just a runtime and not have any starter project layouts.

I can update this PR to do that.

Just a suggestion:
How about `cargo sub.new` for a new submillisecond project?

Starting a Phoenix project is just `mix phx.new`.

PR #182, opened 2023-01-17
Refactor CI workflow

shamilsan:
- Updated `actions/checkout` to v3
- Removed `::set-output` as deprecated (see https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
- Added a command to generate release notes from `CHANGELOG.md`

Release action test: https://github.com/shamilsan/lunatic/actions/runs/3942782116
Release published: https://github.com/shamilsan/lunatic/releases/tag/v0.12.1 (from changelog: https://github.com/shamilsan/lunatic/blob/ebf35ea/CHANGELOG.md)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,10 +84,10 @@ jobs:
       - name: Generate release notes
         if: startsWith(github.ref, 'refs/tags/')
         run: |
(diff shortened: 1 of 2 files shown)


bkolobara:
Thanks for the PR @shamilsan!

@tqwewe also suggested to use a standard format for commits and PRs, so that we could also construct the changelog automatically. I think this would be a good next step, building on top of this to completely automatise the process.

PR #190, opened 2023-02-08
Cli login and deploy

SquattingSocrates:
This code cannot access the `~` directory:

```rust
#[lunatic::main]
fn main(mb: Mailbox<()>) {
    let config = lunatic::ProcessConfig::new().expect("should create new config");
    config.preopen_dir("~");
    spawn_link!(&config, || {
        // this process fails to access home dir
        std::fs::read_dir("~").expect("failed to read home dir");
        println!("Not happening");
    });
}
```


withtypes:
Overall it looks good to me. Can we add a README section which explains how to use it?

SquattingSocrates:
Sure, I'll add a README, no problem.

kosticmarin:
@SquattingSocrates Can you add the README, I'm having trouble testing the new cli.

This is what I managed to test.
Note that I didn't create a `lunatic.toml` not sure how this it is supposed to look when starting a new project.
I've tried by creating an empty `lunatic.toml` and running the following

1. `cargo new cli-test` start new project
2. `lunatic init` init the runner to use lunatic and compile to wasm
3. `touch lunatic.toml`
4. `lunatic auth -p https://localhost`  run auth against local provider

By following the CLI login link I was presented with a prefilled form which is not editable I accept and auth redirects me to `/cli/success` but now my `lunatic.toml` looks like this.

```
project_id = ""
project_url = ""
project_name = ""
provider = ""
remote = []
```
 5. `lunatic app list` Error: Failed to find authenticated provider

Probably I'm missing an important step :smile:

SquattingSocrates:
Hi @kosticmarin! You did everything correctly, but you need to provide the project details. There's a command for setting the current project, listing apps and adding apps etc. I'll get the README ready by the evening today.

But for now, here's an example of how the `project_*` properties look like:
```toml
project_id = "2"
project_url = "project/2/env/2"
project_name = "some new project"
```

src/mode/auth.rs
```diff
@@ -0,0 +1,119 @@
...
+
+async fn check_auth_status(status_url: &str, client: &reqwest::Client) -> Vec<String> {
+    loop {
+        match client.get(status_url).send().await {
+            Ok(res) => {
+                if res.status() == StatusCode::OK {
+                    return res
+                        .cookies()
+                        .map(|cookie| cookie.value().to_string())
+                        .collect();
```
kosticmarin, line 98: Hmm think we should store the whole cookie not just its `value` which is `session_id`.
See [CookieStore::save_json](https://docs.rs/reqwest_cookie_store/0.5.0/reqwest_cookie_store/struct.CookieStore.html#method.save_json) I always get 401 from the api since the cookie is not set properly.

`DEBUG cookie_store::cookie_store] unable to store Set-Cookie: Expired`
`"set-cookie": "lunatic_auth=; Path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT"} }`

PR #192, opened 2023-02-20
Add tests for HashMapId

sid-707:


bkolobara:
Thanks for adding a few tests around the `HashMapId` @sid-707!

PR #194, opened 2023-02-28
Draft: Global process registry - resolves #127

### LINKED ISSUE #127: Global process registry
We should have a global process registry stored in the control node. The API should be similar to the node-local registry.

Yoric:
Well, here is a first draft. I'm fairly confident that I completely misunderstood the task at hand, but this will serve as a basis for further conversations :)

### DIFF
--- a/crates/lunatic-control-axum/src/api.rs
+++ b/crates/lunatic-control-axum/src/api.rs
@@ -119,6 +119,7 @@ impl IntoResponse for ApiError {
     }
 }
 
+impl std::error::Error for ApiError { }
 pub struct JsonExtractor<T>(pub T);
 
 #[async_trait]
--- a/crates/lunatic-control-axum/src/api.rs
+++ b/crates/lunatic-control-axum/src/api.rs
@@ -27,6 +27,7 @@ pub enum ApiError {
     InvalidData(String),
     InvalidPathArg(String),
     InvalidQueryArg(String),
+    ProcessNotFound,
     Custom {
         code: &'static str,
         message: Option<String>,
@@ -42,6 +43,7 @@ impl ApiError {
             ApiError::InvalidData(_) => "invalid_data",
             ApiError::InvalidPathArg(_) => "invalid_path_arg",
             ApiError::InvalidQueryArg(_) => "invalid_query_arg",
+            ApiError::ProcessNotFound => "process_not_found",
             ApiError::Custom { code, .. } => code,
         }
     }
@@ -51,6 +53,7 @@ impl ApiError {
             ApiError::Internal => "".into(),
             ApiError::NotAuthenticated => "Not authenticated".into(),
             ApiError::NotAuthorized => "Not authorized".into(),
+            ApiError::ProcessNotFound => "No such process".into(),
             ApiError::InvalidData(msg) => msg.clone(),
             ApiError::InvalidPathArg(msg) => msg.clone(),
             ApiError::InvalidQueryArg(msg) => msg.clone(),
@@ -106,6 +109,7 @@ impl IntoResponse for ApiError {
             Self::Internal => S::INTERNAL_SERVER_ERROR,
             Self::NotAuthenticated => S::UNAUTHORIZED,
             Self::NotAuthorized => S::FORBIDDEN,
+            Self::ProcessNotFound => S::NOT_FOUND,
             InvalidData(_) | InvalidPathArg(_) | InvalidQueryArg(_) | Custom { .. } => {
                 S::BAD_REQUEST
             }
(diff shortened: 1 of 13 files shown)


Yoric:
After discussing this on Discord, I have a clearer idea of the expected API, so here is a v2.

Not sure how/where I should write tests.

crates/lunatic-control-axum/src/api.rs
```diff
@@ -42,6 +43,7 @@ impl ApiError {
             ApiError::InvalidData(_) => "invalid_data",
             ApiError::InvalidPathArg(_) => "invalid_path_arg",
             ApiError::InvalidQueryArg(_) => "invalid_query_arg",
+            ApiError::ProcessNotFound => "process_not_found",
```
withtypes, line 46: These are more HTTP errors, so we could simply add `NotFound(String)` and put info "Process not found" in it. It's already obvious from the context what HTTP 404 means. Maybe we don't even need the message, just `NotFound`

crates/lunatic-control-axum/src/routes.rs
```diff
@@ -50,6 +51,9 @@ pub async fn register(
             get_module: format!("http://{host}/module/{{id}}"),
             add_module: format!("http://{host}/module"),
             get_nodes: format!("http://{host}/nodes"),
+            get_process: format!("http://{host}/process/{{name}}"),
```
withtypes, line 54: By the way, we should check how path parms in axum work regarding URL encoding https://en.wikipedia.org/wiki/URL_encoding and also names cannot contain / etc. hmm

crates/lunatic-control-axum/src/routes.rs
```diff
@@ -3,13 +3,14 @@ use std::sync::Arc;
...
     extract::DefaultBodyLimit,
-    routing::{get, post},
+    routing::{get, post, delete},
     Extension, Json, Router,
 };
 use lunatic_distributed::{
     control::{api::*, cert::TEST_ROOT_CERT},
     NodeInfo,
 };
+use lunatic_process::{ProcessName, ProcessRecord};
```
withtypes, line 13: Hmm not sure we should require lunatic-process crate just to get a string and node/process id. I think it's ok for the API to simply use String and define it's own {node_id, process_id} struct.

withtypes: Actually, we have merged submilisecond impl of ctrl srv which has lunatic-control crate with only types. We could add those types there...

crates/lunatic-distributed-api/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "Apache-2.0/MIT"
 
 [dependencies]
 lunatic-common-api = { workspace = true }
+lunatic-control-axum = { workspace = true }
```
withtypes, line 12: No, no, if you need some types that are used both in control axum and here, now you have lunatic-control which should keep types in-common.

crates/lunatic-distributed-api/src/lib.rs
```diff
@@ -498,3 +503,153 @@ where
...
+    Box::new(async move {
+        // Extract process name from memory.
+        let memory = get_memory(&mut caller)?;
+        let (memory_slice, state) = memory.data_and_store_mut(&mut caller);
+        let name = memory_slice
+            .get(name_str_ptr as usize..(name_str_ptr + name_str_len) as usize)
+            .or_trap("lunatic::distributed::get")?;
+        let name = std::str::from_utf8(name).or_trap("lunatic::distributed::get")?;
+
+        // Sanity check
```
withtypes, line 530: ?

withtypes (commented):
Starting to look good! Nice work! I left some remarks and questions.

crates/lunatic-control-axum/src/api.rs
Yoric: I'm going to go with no-arg `NotFound` for the time being.

crates/lunatic-control-axum/src/routes.rs
Yoric: I *think* that it can contain `/` as long as it's properly url-encoded.

I'll try and write a test for that.

crates/lunatic-distributed-api/Cargo.toml
Yoric: The reason for this dependency is so that I can use `ApiError`. Should I move `ApiError` to lunatic-control? If so, I suspect that I will need to introduce from lunatic-control to axum, is that alright?

withtypes: I think you don't need them. ApiError is an internal thing, the external interface is HTTP. In the HTTP client you can just match code strings  and HTTP statues which are meaningful to handle. Otherwise, you just report error.

scutuatua-crypto (approved):

PR #195, opened 2023-02-28
Update link for Project Loom

sid-707:
Update the link to use HTTPS so that the browser doesn't give you a warning.


bkolobara:
Thanks for fixing this!

PR #202, opened 2023-03-28
feat: add opentelemetry tracing and metrics

tqwewe:
Todo:
- [x] Each process should have its own single context.
- [x] Ensure doc comments are written correctly.
- [x] Make sure host function trait bounds are only whats required.
- [x] Figure out how to add the environment id to prometheus `target_info`. (Turns out its not possible)
- [x] ~Add push and take functions for resource sharing. Though, this should not "move" the resource, probably would be better to allow multiple processes to share metric resources.~ I don't think this is needed in this PR for now.
- [ ] What do I do with distributed state?

Spans cannot be shared across processes, as they are in a tree structure, and sharing them means it would be possible to drop a parent span before its child, which wouldn't make sense.

_Running the spawn process benchmark, this PR does not seem to affect performance of spawning processes._

Related PRs:
https://github.com/open-telemetry/opentelemetry-rust/pull/1009
https://github.com/open-telemetry/opentelemetry-rust/pull/1018

---

Screenshots for examples/metrics.rs

https://github.com/lunatic-solutions/lunatic-rs/blob/4681561eb78d1164bc1b2eef7c436bcab36622ab/examples/metrics.rs#L21-L78

**Terminal**
```
[2023-04-06T07:22:53Z INFO  metrics] Additional log message, with formatting!
[2023-04-06T07:22:53Z INFO  metrics] formatted object
[2023-04-06T07:22:53Z INFO  metrics] debug object
[2023-04-06T07:22:53Z INFO  metrics] person object
[2023-04-06T07:22:53Z INFO  my_app] a log from my_app
[2023-04-06T07:22:53Z INFO  metrics] a log under my_span
```

**Jaeger**
jaeger

**Prometheus**
prometheus


src/state.rs
```diff
@@ -456,39 +614,61 @@ impl DistributedCtx<LunaticEnvironment> for DefaultProcessState {
...
             registry_atomic_put: None,
+            tracer: Arc::new(BoxedTracer::new(Box::new(NoopTracer::new()))),
+            tracer_context: Arc::new(Context::new()),
+            process_context: Context::new(),
+            meter_provider: GlobalMeterProvider::new(NoopMeterProvider::new()),
+            logger: Arc::new(
+                env_logger::Builder::new()
+                    .filter_level(log::LevelFilter::Off)
+                    .build(),
+            ),
```
tqwewe, line 656: I haven't fully understood about new_dist_state, and have made all logging for this be noops.

Is there any advice on what I should do here? Should logging to the terminal print on the control server?

Cargo.toml
```diff
@@ -127,6 +130,7 @@ anyhow = "1.0"
 bincode = "1.3"
 dashmap = "5.4"
 log = "0.4"
+opentelemetry = { version = "0.19", git = "https://github.com/tqwewe/opentelemetry-rust", branch = "cow", features = ["metrics"] }
```
tqwewe, line 133: Unfortunately this PR uses some changes in opentelemetry-rust which are not yet published.

The two PR's are:
https://github.com/open-telemetry/opentelemetry-rust/pull/1009
https://github.com/open-telemetry/opentelemetry-rust/pull/1018

src/state.rs
withtypes: So dist state only has references to control client and "node". Node implements communication with other nodes (for example, it receives a spawn command from another node). Control client talks to the control server.

We can leave it out for now.

crates/lunatic-metrics-api/src/lib.rs
```diff
@@ -1,161 +1,750 @@
...
+    } else {
+        None
+    };
+
+    let name = get_string_arg(data, name_ptr, name_len).or_trap("lunatic::metrics::span_start")?;
+    let attributes = if attributes_len > 0 {
+        let attributes_data = data
+            .get(attributes_ptr as usize..(attributes_ptr + attributes_len) as usize)
+            .or_trap("lunatic::metrics::span_start")?;
+        let attributes_json =
```
withtypes, line 122: Should vm include some attrs like env/proc id? We should make sure guests cannot override them. This is useful in a multi-tenant/hosted case where don't trust guests.

Cargo.toml
withtypes: I saw PRs got merged, do they plan to release the newer version soon?

withtypes (commented):
Overall it looks good! All metrics include some attributes like: node id, environment id, process id. They are not implicitly set, right? We have to set them on each call? We also want to set them in vm, no guest, so that it can be trusted.

Cargo.toml
tqwewe: Looking at their previous releases, it seems they don't push releases very frequently :(

I've just made a discussion on it, hopefully we can get more insight there
<https://github.com/open-telemetry/opentelemetry-rust/discussions/1031>

tqwewe:
> Overall it looks good! All metrics include some attributes like: node id, environment id, process id. They are not implicitly set, right? We have to set them on each call? We also want to set them in vm, no guest, so that it can be trusted.

Actually the process_id and environment_id is not attached to every trace/metric, but is only attached to the parent spans.
This might be a problem though since if the runtime exits without closing a span properly (which is quite likely), then we might not know which environment the spans come from.

I'll work on injecting this data to every span/log/metric.

crates/lunatic-metrics-api/src/lib.rs
tqwewe: Yep I'll work on adding this

tqwewe: Added this in https://github.com/lunatic-solutions/lunatic/pull/202/commits/4019049d31bd4d2a8c75f9875a9b1be243d64e8f

scutuatua-crypto (approved):

PR #210, opened 2023-05-23
Fix lookup nodes host call

kosticmarin:
Since we transitioned to a HTTP based control server we were missing filtering nodes based on node attributes defined by `--tag` when running `lunatic node`.

This is done via `lunatic::distributed::lookup_nodes` host function.

There is one little overhead, I've noticed that we keep node attributes as json `Value` (probably sqlite related), so each request needs to de-serialize attributes when filtering nodes. Not sure if we want to change this?

Also I've bumped the `lunatic-control-submillisecond` deps.

### DIFF
--- a/src/mode/node.rs
+++ b/src/mode/node.rs
@@ -62,7 +62,7 @@ pub(crate) async fn start(args: Args) -> Result<()> {
     // TODO unwrap, better message
     let node_name = Uuid::new_v4();
     let node_name_str = node_name.as_hyphenated().to_string();
-    let node_attributes: HashMap<String, String> = Default::default(); //args.tag.into_iter().collect(); TODO
+    let node_attributes: HashMap<String, String> = args.tag.clone().into_iter().collect();
     let node_cert = lunatic_distributed::distributed::server::gen_node_cert(&node_name_str)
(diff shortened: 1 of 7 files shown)


withtypes:
Let's change it so that value of attrs is string, not json value. Not sure when or why it became json value.

src/mode/node.rs
```diff
@@ -62,7 +62,7 @@ pub(crate) async fn start(args: Args) -> Result<()> {
     // TODO unwrap, better message
     let node_name = Uuid::new_v4();
     let node_name_str = node_name.as_hyphenated().to_string();
-    let node_attributes: HashMap<String, String> = Default::default(); //args.tag.into_iter().collect(); TODO
+    let node_attributes: HashMap<String, String> = args.tag.clone().into_iter().collect();
```
tqwewe, line 65: Everything looks good to me. I don't have my head around the entire control server much to spot any issues.
In terms of the `BincodeJsonValue`, I'm not sure where that came from as I don't recall adding it, I mostly just copied from the axum implementation.

Here I noticed though, from a glance this `.clone()` might not be necessary.

PR #211, opened 2023-06-28
Privileged and Unprivileged Nodes

teskeras:


crates/lunatic-process/src/env.rs
```diff
@@ -111,12 +111,12 @@ pub struct LunaticEnvironments {
 #[async_trait]
 impl Environments for LunaticEnvironments {
     type Env = LunaticEnvironment;
-    async fn create(&self, id: u64) -> Arc<Self::Env> {
+    async fn create(&self, id: u64) -> Result<Arc<Self::Env>> {
```
withtypes, line 114: Why result?

kosticmarin:
This PR adds definitions of "privileged" and "unprivileged" nodes.

Privileged nodes can process distributed messages from any environment, while unprivileged nodes on registration receive a list of allowed environments. Any message that an unprivileged node process and which comes from an environment to which the node does not have access will result in an error.

By default control servers implementations, provided in this repository (see crates `lunatic-control-axum` and `lunatic-control-submilliseconds`), register only privileged nodes with no limitations.

PR #217, opened 2023-08-23
chore: Remove macOS universal build

kosticmarin:
- Release separate binaries for x86_64 and aarch64

This was done in order to fix issues with existing universal binary which does not work on some macOS systems.
Some people reported the following issue `exec format error: lunatic`.

@bkolobara I'm not sure how to test this without creating a tag?


bkolobara:
I never figured out a way to test this without actually pushing to the repository.

Creating tags should be fine, as it only creates a "draft release" that we don't need to publish. I think you should be able to see draft releases with your repository permissions.

kosticmarin:
@bkolobara Every thing looks good! Can You try out the new binaries on macOS, to see if the problems format are gone?

After we can release the draft builds.

PS. Ignore the deny checks :smile:

bkolobara:
It works!

But it also bundles the whole `target` directory structure. We should probably just include the binary at the top level, like on linux and windows.

I just remembered that we also have a [hombrew formula](https://github.com/lunatic-solutions/homebrew-lunatic/blob/main/Formula/lunatic.rb) that is [automatically updated](https://github.com/lunatic-solutions/homebrew-lunatic/blob/main/.github/workflows/ci.yml#L18-L26) when a release [is triggered](https://github.com/lunatic-solutions/lunatic/blob/main/.github/workflows/homebrew.yml). I couldn't find examples of formulas that have both arm and x64 binaries in it. With the universal build we always pointed to the same, but am unsure how it would work now.

kosticmarin:
Ok, great catch I'll remove the directory structure!

Also I'll check what we can do about hombrew and how we can improve release there.

kosticmarin:
I've fixed the directory structure when we package macOS builds.

Also I've checked some of the most popular rust project and how they are installed using Homebrew. Most of them are using `cargo` and installing from source via-a github release archive.

Here are some of the examples:

- [wasmtime.rb](https://github.com/Homebrew/homebrew-core/blob/f981b61d166730ed342b85eab5f4b85dfc65177c/Formula/w/wasmtime.rb)
- [bat.rb](https://github.com/Homebrew/homebrew-core/blob/f981b61d166730ed342b85eab5f4b85dfc65177c/Formula/b/bat.rb)
- [ripgrep.rb](https://github.com/Homebrew/homebrew-core/blob/f981b61d166730ed342b85eab5f4b85dfc65177c/Formula/r/ripgrep.rb)

In short all their formula depend on `depends_on "rust" => :build` and they just run `system "cargo", "install"`.

scutuatua-crypto (approved):

